PR #3, opened 2017-01-09
fix port in README

chanezon:
port 3000 -> 8080


alexellis:
Oops.. I had to change to port 3000 on the machine I tested on, thanks for the help.

PR #10, opened 2017-01-23
Better debugging infos

thejibz:
When the "fprocess" started by the watchdog is failing, the stderr of the "fprocess" is lost... That made debugging very hard cause you got output from the "panic(err)" whitout much infos like this : "2017/01/23 10:55:20 http: panic serving 10.0.1.3:41882: exit status 1".

With this change, you have also the stderr of the "fprocess" printed in console (docker logs [container_id]).

### DIFF
--- a/watchdog/main.go
+++ b/watchdog/main.go
@@ -33,7 +33,7 @@ func main() {
 			fprocessCmd := exec.Command(parts[0], parts[1:]...)
 			//
 
-			// get a reference on all outputs for the fprocess (as byte[])
+			// get a reference on all outputs for the fprocess (as bytes.Buffer)
 			var stdout bytes.Buffer
 			fprocessCmd.Stdout = &stdout
 
--- a/watchdog/main.go
+++ b/watchdog/main.go
@@ -14,8 +14,8 @@ import (
 func main() {
 	s := &http.Server{
 		Addr:           ":8080",
-		ReadTimeout:    25 * time.Second,
-		WriteTimeout:   25 * time.Second,
+		ReadTimeout:    5 * time.Second,
+		WriteTimeout:   5 * time.Second,
 		MaxHeaderBytes: 1 << 20,
 	}
 
--- a/watchdog/main.go
+++ b/watchdog/main.go
@@ -27,9 +27,9 @@ func main() {
 			res, _ := ioutil.ReadAll(r.Body)
 			writer.Write(res)
 			writer.Close()
-			out, err := targetCmd.Output()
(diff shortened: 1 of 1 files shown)


thejibz:
Hello thanks for the quick reply.
I will follow your recommendation regarding repo and fwatchdog.

Concerning the change, in my personal use case I want to have the error reported to the caller (through the POST body response). Maybe with the gateway/fwatchdog returning a HTTP/500 containing the error msg. I will try to roll out a change in this direction (I am a begginer in Go so it might take a little while). 

I see two cases to keep in mind:
1. fprocess exit status code != 0
  => HTTP/500 + erros msg
2. fprocess exit status code == 0 && stderr not empty (partial failure)
  => Just logging to the stdout of fwatchdog ? HTTP/500 ?

Also, I wonder if the fwatchog should always "panic" in case of a fprocess error. In some instance, it can be related to only a specific input and its not worth shutdown the container all together. But this feature would involve tracking a sort of ratio success/failure because if all the input cause error then restarting the container is indeed a good measure... Sorry, looks like my mind start to drift xD I see I am beginning to suggest implementing a circuit breaker pattern...

PS:  Many thanks for sharing your amazing work !

alexellis:
Thanks for this input, let's move discussion over to an issue and plan a bit more?

alexellis:
Hey @thejibz did you see my comment from earlier?

> Thanks for this input, let's move discussion over to an issue and plan a bit more?

Can we move the discussion over to the issue and decide on direction?

https://github.com/alexellis/faas/issues/11

I also pushed a small change to the labels_metrics branch that may help with your scenario and re-released a 0.2 version with a new binary.

You can add it to containers with an `ADD https://` command.

thejibz:
Hey @alexellis , yes I saw your comment to plan a bit more. But I also saw your other [request to test your code](https://github.com/alexellis/faas/commit/502e13acab742efb1886e9dbb83c0096e74d6280#commitcomment-20579157). I though it was a good idea to amended this pull request with the result of my review. Let's say it's a PoC helping to plan out the feature.
I will check out your change made on labels_metrics and fuel the feature's discussion.

thejibz:
I close this pull request. A new one might be open when the feature's planning will be more advanced.

alexellis:
Yes - I wanted to know whether this addressed your use-case:

https://github.com/alexellis/faas/commit/b7d2845d6e10c0bf422226373949b6e861e56ba3

thejibz:
Yes, thank you it was helpful to fulfill my need.

PR #18, opened 2017-03-03
fixed typo in gateway Dockerfile

### LINKED ISSUE #17: UI portal on 8080 only white screen
This way cool, and thanks for posting this project!

I ran the quickstart, and again on my local machine, however I wasn't able to successfully bring the UI portal up.  
1) Cloned mater branch
2 ) `docker swarm init --advertise-addr=$(ifconfig eth0| grep 'inet addr:'| cut -d: -f2 | awk '{ print $1}') &&   git clone https://github.com/alexellis/faas &&   cd faas &&   ./deploy_stack.sh &&   docker service ls`
3) used curl to verify the stack was consumer web events
4) checked prom to which came up without issues
5) checked  AlertManager also no issues.
6) tried http://localhost:8080 and I can only see a white screen, even though I can post events.

Curious if you have any tips to get to the UX screen.
TIA!

macneib:
fixes #17  typo in Dockerfile

### DIFF
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -8,6 +8,6 @@ ENV https_proxy     ""
 
 COPY gateway    .
 
-COPY assets     assets
+COPY assets     assets/
 
 CMD ["./gateway"]


alexellis:
`COPY x x` is a valid syntax. I think the issue was that I used Docker Cloud for an auto-build yesterday and ended up replacing the wrong image tag. I could reproduce the error and tested the fix so closing this PR.

PR #61, opened 2017-04-20
Switch from quay.io

alexellis:
quay.io and Docker hub images are identical. Switching over to the Hub since quay seems to have been blocked in China for use on Alibaba cloud.


alexellis:
Needs testing / running through locally before switching over.

alexellis:
Fixed by moving images to hub builds.

PR #75, opened 2017-05-02
updated sentiment analysis function to handle multiple sentences and …

aafrey:
…fix decoding issue

Include the changes that Alex had made to the Sentiment Analysis function to better handle multiple sentences. Also includes a fix for certain decoding issues

## Description

Handles multiple sentences and sets the default encoding to UTF-8 to better handle certain characters.
## Motivation and Context

Adds ability to handle multiple sentences as well as fix issues handling certain ASCII characters

## How Has This Been Tested?

Tested by forwarding tweets of a certain hashtag to the function.

Added the function to a FaaS instance running on Vultr. 

They do not impact any code outside of the sentiment analysis function

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] My code follows the code style of this project.
- [ x] All new and existing tests passed.

### DIFF
--- a/sample-functions/SentimentAnalysis/handler.py
+++ b/sample-functions/SentimentAnalysis/handler.py
@@ -1,17 +1,32 @@
 import sys
+import json
 from textblob import TextBlob
+
+# set default encoding to UTF-8 to eliminate decoding errors
+reload(sys)
+sys.setdefaultencoding('utf8')
 
 def get_stdin():
     buf = ""
     for line in sys.stdin:
         buf = buf + line
-    return buf
+        return buf
 
 if(__name__ == "__main__"):
     st = get_stdin()
     blob = TextBlob(st)
-    out =""
-    for sentence in blob.sentences:
-        out = out + "Polarity: " + str(sentence.sentiment.polarity) + " Subjectivity: " + str(sentence.sentiment.subjectivity)  + "\n"
-    print(out)
+    res = {
+        "polarity": 0,
+        "subjectivity": 0
+    }
 
+    for sentence in blob.sentences:
+        res["subjectivity"] = res["subjectivity"] + sentence.sentiment.subjectivity
+        res["polarity"] = res["polarity"] + sentence.sentiment.polarity
+
+    total = len(blob.sentences)
+
+    res["sentence_count"] = total
+    res["polarity"] = res["polarity"] / total
+    res["subjectivity"] = res["subjectivity"] / total
+    print(json.dumps(res))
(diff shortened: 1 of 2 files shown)


sample-functions/SentimentAnalysis/handler.py
```diff
@@ -1,17 +1,31 @@
...
-    out =""
+    res = {
+        "polarity": 0,
+        "subjectivity": 0
+    }
+
     for sentence in blob.sentences:
-        out = out + "Polarity: " + str(sentence.sentiment.polarity) + " Subjectivity: " + str(sentence.sentiment.subjectivity)  + "\n"
-    print(out)
+        res["subjectivity"] = res["subjectivity"] + sentence.sentiment.subjectivity
```
developius: This (and the one below) can be changed to this to follow DRY

```python
res["subjectivity"] += sentence.sentiment.subjectivity
res["polarity"] = += sentence.sentiment.polarity
```

You can also use `/=` for division:

```python
res["polarity"] /= total
res["subjectivity"] /= total
```

alexellis:
Let's keep the text more verbose for readability. Will look at the dry comment later.

sample-functions/SentimentAnalysis/handler.py
```diff
@@ -1,17 +1,31 @@
 import sys
+import json
 from textblob import TextBlob
 
+reload(sys)
```
alexellis: What is `reload` for.. I don't think I've seen it before?

alexellis:
Thanks for this PR @aafrey

sample-functions/SentimentAnalysis/handler.py
aafrey: the setdefaultencoding method does not exist at this point in the program, I needed to reload the sys module for it to be available. Reason being I was seeing a ton of decoding errors when piping in tweets that had odd characters. Setting the default encoding to UTF-8 instead of ASCII solved those issues for me, although maybe there is a better way? I wouldn't think this would be necessary for all python functions, just use cases like this, where unexpected, odd characters can pop up.

alexellis: OK - if you could add a one-liner to explain that I will merge.

aafrey: do you need a one liner in the commit message? or so you want it in a code comment?

alexellis: Above the reload call

PR #79, opened 2017-05-05
Add windows function

StefanScherer:
Add some `node.platform.OS` constraints to allow faas to run in a cross-platform swarm.
Run a first Windows function.

## How Has This Been Tested?
Create a cross-platform Docker swarm with eg. https://github.com/StefanScherer/docker-windows-box/tree/master/swarm-mode

## Types of changes

faas-windows

### DIFF
--- a/docker-compose.windows.yml
+++ b/docker-compose.windows.yml
@@ -1,0 +1,25 @@
+version: "3"
+services:
+
+    # Calculates base64 representation of request body.
+    windows:
+        image: stefanscherer/base:golang-1.7.5-windows
+        labels:
+            function: "true"
+        depends_on:
+            - gateway
+        networks:
+            - functions
+        environment:
+            fprocess: "app.exe"
+            no_proxy: "gateway"
+            https_proxy: $https_proxy
+        deploy:
+            placement:
+                constraints: [node.platform.OS == windows]
+
+networks:
+    functions:
+        driver: overlay
+        # Docker does not support this option yet - maybe create outside of the stack and reference as "external"?
+        #attachable: true
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
         deploy:
             placement:
                 constraints: [node.role == manager]
- 
+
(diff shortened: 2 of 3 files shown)


amirmc:
Nice!

alexellis:
Thanks for this Stefan - the last piece in the puzzle while trying to get Docker + Windows 2016 working.

> Something like? Alex’s FaaS platform runs on any cloud and lets you write functions in any programming language mixing Windows and Linux. 👍

vielmetti:
Nice work. This design should also work for constraining to linux/arm64 vs linux/armhf vs linux/amd64 ?

StefanScherer:
@vielmetti Yes, this PR https://github.com/docker/swarmkit/pull/1678 has introduced it: Eg. `node.platform.os == linux`, `node.platform.arch == x86_64`
Wow, the arch will be the more difficult one to remember, x86_64 or amd64 ? I think of the manifest-tool where ARCH is amd64 😁

StefanScherer:
Just to complete the `node.platform.arch` topic with my comment https://github.com/moby/moby/pull/27455#issuecomment-299518582 it seems we have to live with both in a multi-arch swarm running multi-arch images:

```
$ docker image inspect alpine | grep Arch
        "Architecture": "amd64",

$ docker node inspect moby | grep Arch
                "Architecture": "x86_64",
```

alexellis:
I don't think I can merge the windows service into the compose file since it will mostly be deployed to a Linux-only cluster. This is good reference though. :+1:

StefanScherer:
@alexellis I've splitted the windows service into a different yml file. You can cherry-pick the commits you like to have in your repo or now merge it. But even with a single docker-compose.yml there was no problem on a Linux only swarm as that service can be created, but keeps at replicas 0/1.

alexellis:
I'll close the PR, but I've tracked the number for inclusion at a later stage. Since the primary target is Linux I'll leave out the additional Node OS restrictions and add them to the documentation later. Thanks again.

PR #84, opened 2017-05-18
Add try in pwd button.

marcosnils:
## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ A one-line script is provided to help you get started quickly. You can test-driv
 
 Use your own laptop or the free community-run Docker playground: play-with-docker.com.
 
+[![Try in PWD](https://cdn.rawgit.com/play-with-docker/stacks/cff22438/assets/images/button.png)](http://play-with-docker.com?stack=https://raw.githubusercontent.com/alexellis/faas/master/docker-compose.yml)
+
 ### [Begin the TestDrive](https://github.com/alexellis/faas/blob/master/TestDrive.md)
 
 Here is a screenshot of the API gateway portal - designed for ease of use.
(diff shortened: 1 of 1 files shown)


alexellis:
This is cool.

Would be good to merge when 
- the stack name can match one given in the link (so that the test drive instructions match with the deploy_stack.sh function names)
- and the Safari scroll issue is fixed

marcosnils:
Done @alexellis. /cc @alexellis

alexellis:
Great! I'll try it out.

developius:
After deployed to PWD, I can't access ports 9090 & 9093 - any reason for this? 8080 works fine for the UI.

alexellis:
How did you deploy it to PWD - using the new button from @marcosnils or the TestDrive/script way ?

alexellis:
@marcosnils only fails of deployed with the new button?

marcosnils:
@alexellis if I either deploy it using the button or copying the docker-compose file and doing `docker stack deploy` the result is the same. Only 8080 is accessible. Not sure how 9090 and 9093 should be used though. The "Try in PWD"  does the same thing as if you deploy it manually with `docker stack deploy`

marcosnils:
Another thing that I've noticed is that after deploying if I do `curl localhost:9090` or `curl localhost:9093` it doesn't work (which makes sense). I don't think it's a PWD issue.

developius:
@alexellis I used the button. @marcosnils 9090 & 9093 are for Prometheus - they worked for me when I did a `docker stack deploy` on my own swarm.

alexellis:
Opening 9090 in a browser worked for me yesterday with the TestDrive instructions. Also works on VMs.

Issue #83, opened 2017-05-18
Support private Docker registry


sebgl:
When using our own private docker registry, we need the swarm cluster to be able to pull images (functions) from it. 

## Expected Behavior
When a new function service is created in Swarm, images from a private registry should be correctly pulled provided the registry credentials are available somewhere.

## Current Behavior
Images cannot be pulled by Swarm workers.

## Possible Solution
We can rely on the mechanism provided by the `--with-registry-auth` flag in the `docker stack` command, which forwards the registry credentials to all swarm nodes.
When creating a service using the golang Docker sdk, we can pass the appropriate encoded registry auth (base64 encoding of the json string representation).
I have a proposal there, based on environment variables configuration: https://github.com/sebgl/faas/commit/e7d2debf122a5132f32023197be03b7a5e83e79a
If you think that makes sense I can open a pull request.

## Steps to Reproduce (for bugs)
1. Push function image to a private Docker registry
2. Try to invoke the function with FaaS (making sure the docker image is not already available locally on swarm nodes)

## Context
We would like to use FaaS with private Docker registries (not only hub.docker.com but any registry).

## Your Environment
* Docker version: Docker 17.05.0-ce

alexellis:
Hi @sebgl - this makes sense and I'd like to support private registries.

Questions:

* Can [swarm secrets](http://blog.alexellis.io/swarm-secrets-in-action/) replace the plaintext environmental variables?

* Could you have to support multiple private registries at once? If that were the case how would that be possible with the proposal?

* Do you have a 1-liner to start a private registry with username/password enabled (TLS?)

sebgl:
Hi @alexellis , thanks for the quick answer.

My use case (having only one single registry) is a tiny subcase of the more general one.
Since this might interest you: we're basically adding another API in front of the gateway to handle authentication, permissions, etc. This API is also responsible for building and pushing docker images, since we want users to give us code, not docker images. Hence we only need one private registry (ours, not the user's one). For now this is just a prototype, for fun :)

You're right in all questions that this should be handled in a more general way in FaaS gateway.

* Multiple private registries
Yes ! Should this be a fixed list of allowed registries? Or should users be able to "add" access to new registries whenever they want via the gateway API (can play nice with registry permissions)? The latter seems more interesting but also a bit more complex to handle. I love how small and simple the codebase of FaaS is.

* Credential store & swarm secrets
[...]

PR #87, opened 2017-05-26
Handle private docker registry auth

sebgl:
This adds support for private docker registries.

## Description
An optional `registryAuth` field in the CreateFunctionRequest
allows users to specify a registry auth to be used for deployment.

Auth must be passed as base64-encoded basic auth, similar to
how it's done in Docker file store credentials (~/.docker/config.json).

Credentials are then passed to swarm at service creation.

## Motivation and Context
Support for private registries.
See issue #83 

## How Has This Been Tested?
Manual test procedure:
1) deploy a private registry with basic auth enabled (see https://gist.github.com/sebgl/3c97c379ddc77c65c44d47e6b745fa10)
2) push a docker image to the private registry
3) deploy a function with `"registryAuth": "base64user:password"`

Automated integration tests are a bit harder to perform due to the
need for a private registry with TLS + basic auth.

## Types of changes

## Checklist:

A few things I'd like to point out:
* The only documentation I have updated is `api-docs/swagger.yml`. I'm not sure that's enough. Should we include additional documentation somewhere?
* In order to stay close to the authentication system set up with Docker credentials (`~/.docker/config.json`), the deploy API accepts a base64-encoded basic auth. This might be a bit confusing.
* As discussed with Alex, the web UI has not be updated to match this change for now.

### DIFF
--- a/api-docs/swagger.yml
+++ b/api-docs/swagger.yml
@@ -155,6 +155,9 @@ definitions:
         items:
           type: "string"
         description: "Overrides to environmental variables"
+      registryAuth:
+        type: "string"
+        description: "Private registry base64-encoded basic auth (as present in ~/.docker/config.json)"
 externalDocs:
   description: "More documentation available on Github"
   url: "http://docs.get-faas.com"
--- a/docs/managing-images.md
+++ b/docs/managing-images.md
@@ -1,0 +1,86 @@
+# Managing images
+
+## Using private Docker registries
+
+FaaS supports running functions from Docker images in private Docker registries.
+The registry credentials can be passed on function deployment, and are then handled by Swarm for image polling.
+
+### Deploy functions with private registries credentials
+
+A `POST` request on `/system/function` allows you to specify private registry credentials, as a base64-encoded basic auth (user:password).
+```
+curl -XPOST /system/functions -d {
+    "service": "functionName",
+    "image": "privateregistry.domain.com/user/function",
+    "envProcess": "/usr/bin/myprocess",
+    "network": "func_functions",
+    "registryAuth": "dXNlcjpwYXNzd29yZA=="
+}
+```
+
+Base64-encoded basic auth can be resolved using your registry username and password:
+````
+echo -n "user:password" | base64
+````
+
+You can also find it in your `~/.docker/config.json` Docker credentials store, as a result of the `docker login` command:
+```
+cat ~/.docker/config.json
+{
+	"auths": {
+		"privateregistry.domain.com": {
+			"auth": "dXNlcjpwYXNzd29yZA=="
+		}
+	}
+}
+```
+
+### Deploy your own private Docker registry
+
+If you wish to deploy your own private registry, you can follow [Docker official documentation](https://docs.docker.com/registry/deploying/).
+
+A quick way to get started for a private registry with TLS and authentication
+is to create a VM with port 443 open to the world (for letsencrypt registration), and a registered DNS ($YOURHOST).
+Then, create these two files in the current directory:
+
+```
+# docker-compose.yml
+version: '2'
+
+services:
+
+  registry:
+    restart: always
+    image: registry:2
+    ports:
+      - 5000:5000
+      - 443:5000
+    environment:
+      REGISTRY_AUTH: htpasswd
+      REGISTRY_AUTH_HTPASSWD_PATH: /auth/htpasswd
+      REGISTRY_AUTH_HTPASSWD_REALM: Registry Realm
+      REGISTRY_HTTP_TLS_LETSENCRYPT_CACHEFILE: /letsencrypt/cache
+      REGISTRY_HTTP_TLS_LETSENCRYPT_EMAIL: your@email.com
+    volumes:
+      - ./data:/var/lib/registry
+      - ./auth:/auth
+      - ./letsencrypt:/letsencrypt
+```
+
+```
+# auth/htpasswd (generated with `docker run --entrypoint htpasswd registry:2 -Bbn testuser testpassword`)
+testuser:$2y$05$Bl9siDMe7ieQHLM8e7ifaOklKrHmXymbMqfmqXs7zssj6MMGQW4le
+```
+
+Your registry is ready to be deployed by running `docker-compose up -d`.
+
+On the client machine, you can now login and use the newly setup registry:
+```
+docker pull ubuntu && docker tag ubuntu $YOURHOST/ubuntu
+docker login $YOURHOST # will add encoded registry credentials to ~/.docker/config.json
+    Username: testuser
+    Password: testpassword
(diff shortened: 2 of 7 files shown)


alexellis:
Testing:

```
docker pull vielmetti/faas-oblique
docker tag vielmetti/faas-oblique domain.com/faas-oblique
docker login domain.com
docker push domain.com/faas-oblique
```

```
curl localhost:8080/system/functions -d '
{"service": "oblique", "image": "push domain.com/faas-oblique", "envProcess": "/usr/bin/oblique", "network": "func_functions", "registryAuth": "base64encodedhere"}' -v -4
```

gateway/handlers/functionshandler.go
```diff
@@ -208,3 +223,43 @@ func makeSpec(request *requests.CreateFunctionRequest) swarm.ServiceSpec {
...
+	c, err := base64.StdEncoding.DecodeString(basicAuthB64)
+	if err != nil {
+		return "", "", err
+	}
+	cs := string(c)
+	s := strings.IndexByte(cs, ':')
+	if s < 0 {
+		return "", "", errors.New("Invalid basic auth")
+	}
+	return cs[:s], cs[s+1:], nil
```
alexellis: We might benefit from some range-checking here.

gateway/handlers/functionshandler.go
```diff
@@ -208,3 +223,43 @@ func makeSpec(request *requests.CreateFunctionRequest) swarm.ServiceSpec {
 
 	return spec
 }
+
+func buildEncodedAuthConfig(basicAuthB64 string, dockerImage string) (string, error) {
```
alexellis: Can we get a unit test against this with some known values?

api-docs/swagger.yml
```diff
@@ -155,6 +155,9 @@ definitions:
         items:
           type: "string"
         description: "Overrides to environmental variables"
+      registryAuth:
+        type: "string"
+        description: "Private registry base64-encoded basic auth"
```
alexellis: Can you mention `~/.docker/config` as a source for the data?

alexellis:
Hi @sebgl thanks a lot for this PR - this looks great and I was able to test before/after with LetsEncrypt on a test domain 🍾 

I think we need a short guide including your instructions from the Gist for people to understand how this works at a high level and how to make use of it.

I've also added a couple of comments to the current PR.

sebgl:
Thanks for the review. I agree with all your comment, hence I will update the PR with everything you suggested. Give me a few days (family week-end :) ).
What's the best place for documentation and instructions? I was thinking about a `registry.md` file at the root of the project?

alexellis:
The docs directory would seem like a good place to put this.

https://github.com/alexellis/faas/tree/master/docs

It is automatically converted to HTML at http://docs.get-faas.com

We can link to a using FaaS or deploying FaaS page from the landing site. On there should be a link to the guide.

Deploying FaaS

* Enabling HTTPs
* Managing FaaS  images / Using private repositories
* Getting the CLI

(etc)

gateway/handlers/functionshandler.go
sebgl: What do you mean? I don't really this a case where this would fail (even with cs == ":").
Part of this code comes from the standard library: https://golang.org/src/net/http/request.go?#L848

gateway/handlers/functionshandler.go
sebgl: Sure !

api-docs/swagger.yml
sebgl: Yes !

gateway/handlers/functionshandler.go
```diff
@@ -208,3 +223,43 @@ func makeSpec(request *requests.CreateFunctionRequest) swarm.ServiceSpec {
...
+	return base64.URLEncoding.EncodeToString(buf), nil
+}
+
+func userPasswordFromBasicAuth(basicAuthB64 string) (string, string, error) {
+	c, err := base64.StdEncoding.DecodeString(basicAuthB64)
+	if err != nil {
+		return "", "", err
+	}
+	cs := string(c)
+	s := strings.IndexByte(cs, ':')
```
alexellis: Curious about `strings.IndexByte` vs `strings.Index`?

sebgl: `strings.IndexByte` is looping over individual bytes (char), for performance reason. It seems to be a subset of `strings.Index` which looks for strings with multiple characters, and actually calls `strings.IndexByte`when the string to look for is a single char. See https://golang.org/src/strings/strings_amd64.go?#L31

sebgl:
Changes pushed:
* commit is signed
* documentation in `docs/managing-images.md`
* mention `~/.docker/config.json` in the swagger schema
* unit tests for `BuildEncodedAuthConfig`

gateway/handlers/functionshandler.go
alexellis: Great, thanks for the explanation. I think I can replace other instances of `string.Index` to use this too.

docs/managing-images.md
```diff
@@ -0,0 +1,81 @@
...
+curl -XPOST /system/functions -d {
+    "service": "functionName",
+    "image": "privateregistry.domain.com/user/function",
+    "envProcess": "/usr/bin/myprocess",
+    "network": "func_functions",
+    "registryAuth": "dXNlcjpwYXNzd29yZAo"
+}
+```
+
+Base64-encoded basic auth can be computed manually, or found in your `~/.docker/config.json` Docker credentials store, as a result of the `docker login`command:
```
alexellis: "computed manually".. an example would be what - `echo "username:password" | base64`?

gateway/handlers/functionshandler.go
alexellis: Sure - OK.

alexellis:
When signed off the commit message gets something added like `Signed-off-by:`.

Do you want to add the unit test execution to the Dockerfile.build.* files `go test -v`? I'm not sure they ran as part of the Travis CI build.

docs/managing-images.md
sebgl: Yes ! I will add the example.

sebgl:
Sorry for the signature, should be fixed now. I also added the base64 example you suggested.

sebgl:
Indeed for the unit tests, sounds like a good idea. I pushed it as a second commit (could be squashed in a single one or proposed in another PR if you prefer).

alexellis:
Hi @sebgl - did you ever start the work for the FaaS CLI to make use of private registries?

sebgl:
Hi @alexellis 
Sorry for the lack of news lately, other priorities came up and I had to move away from playing with FaaS for a while.
I'm not sure I'll have time to get into the FaaS CLI in the near future - I'm sorry about that. If you want to implement the feature I'll be happy to have a look at it though.

PR #93, opened 2017-06-29
Add link to FaaS Figlet sample

jmkhael:
Add link to FaaS Figlet blog post.

1. a light image is built from an Alpine base
2. uses Figlet tool internally
3. demo how to use faas-cli to deploy it.

So that we have a FaaS function generate Ascii banners.


alexellis:
Thanks @jmkhael !

Issue #94, opened 2017-06-29
Functions that run longer than 8 seconds return empty replies.


stealthybox:
Long running functions (8+ seconds) return with empty responses.

## Expected Behavior
STDOUT should be returned in the response body.

## Current Behavior
The response body is empty.

## Steps to Reproduce
Deploy faas.
Create the following files:

#### func.yml
```yaml
version: '3.1'

services:
  test:
    image: functions/alpine
    networks: [ func_functions ]
    labels:
      function: 'true'
    environment:
      read_timeout: 60
      write_timeout: 60
      fprocess: 'sh /root/func.sh'
    volumes:
    -  .:/root
networks:
  func_functions:
    external: true
```
#### func.sh
```bash
DELAY=`cat -`
echo hi, sleeping for $DELAY
sleep $DELAY
echo done sleeping, bye!
```
This function takes a number as input and sleeps for that duration before responding.
Let's deploy it:
```bash
> docker deploy -c func.yml issue
> docker service logs -f issue_test
```
And curl the service to run for different lengths:
```bash
> time curl localhost:8089/function/issue_test -XPOST -d6
hi, sleeping for 6
done sleeping, bye!

real    0m6.013s
user    0m0.004s
sys     0m0.001s

> time curl localhost:8089/function/issue_test -XPOST -d7
hi, sleeping for 7
done sleeping, bye!

real    0m7.013s
user    0m0.000s
sys     0m0.006s

> time curl localhost:8089/function/issue_test -XPOST -d8
curl: (52) Empty reply from server

real    0m8.013s
user    0m0.000s
sys     0m0.005s

> time curl localhost:8089/function/issue_test -XPOST -d9
curl: (52) Empty reply from server

real    0m9.013s
user    0m0.002s
sys     0m0.003s

```
The test cases for 8 and 9 seconds still return a response, but the response is empty.

## Context
@cgpuglie and I are trying to run a web-scraper with nightmare-js as a faas function.
The runtime is variable, but typically longer than 8 seconds.

## Your Environment
```
> docker version
Client:
 Version:      17.05.0-ce
 API version:  1.29
 Go version:   go1.7.5
 Git commit:   89658be
 Built:        Thu May  4 22:10:29 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.05.0-ce
 API version:  1.29 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   89658be
 Built:        Thu May  4 22:10:29 2017
[...]

PR #96, opened 2017-07-02
Gateway Configuration with Timeouts

### LINKED ISSUE #95: Allow configurable HTTP timeout in Gateway
To resolve #94 or provide temporary work-around.

## Expected Behavior

The API Gateway has hard-coded 8-second timeout, this should be configurable. Most people will want this to be shorter - some longer.

## Current Behavior

8-seconds is the max duration

## Possible Solution

Introduce environmental configuration to match that of the watchdog's configuration options.

https://github.com/alexellis/faas/blob/master/watchdog/readconfig.go

See #94 or set `fprocess` to `sleep 10`.

cc/ @stealthybox

stealthybox:
New feature for Gateway Configuration Options

## Description

- Adds GatewayConfig
- Env configurable `read_timeout` & `write_timeout` /w test copied over from watchdog.
- Moves httpPort definition to the config default declaration.
- Modifies Dockerfiles to add new files and test code.
- Adds runtime logging for options.

## Motivation and Context
Resolves #95 and #94 

## How Has This Been Tested?
Unit test and config were included in the `main` package instead of the `tests` folder because of access to unexported struct members.
The additional test gates the build.
This follows the same format as the watchdog.

I tested e2e functionality by setting the `read_timeout` and `write_timeout` ENV vars on the gateway container, checking the new log output, and following the steps in #94.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/Dockerfile.build
+++ b/gateway/Dockerfile.build
@@ -18,8 +18,6 @@ COPY requests       requests
 COPY tests          tests
 COPY server.go      .
 COPY readconfig.go  .
-COPY config_test.go .
 
-RUN go test && \
-    go test -v ./tests && \
+RUN go test -v ./tests && \
     CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .
--- a/gateway/Dockerfile.build
+++ b/gateway/Dockerfile.build
@@ -10,13 +10,16 @@ FROM golang:1.7.5
 
 WORKDIR /go/src/github.com/alexellis/faas/gateway
 
-COPY vendor     vendor
+COPY vendor         vendor
 
-COPY handlers   handlers
-COPY metrics    metrics
-COPY requests   requests
-COPY tests      tests
-COPY server.go  .
+COPY handlers       handlers
+COPY metrics        metrics
+COPY requests       requests
+COPY tests          tests
+COPY server.go      .
+COPY readconfig.go  .
+COPY config_test.go .
 
-RUN go test -v ./tests && \
+RUN go test && \
+    go test -v ./tests && \
     CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .
--- a/gateway/config_test.go
+++ b/gateway/config_test.go
@@ -1,0 +1,61 @@
+// Copyright (c) Alex Ellis 2017. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package main
+
+import (
+	"testing"
+	"time"
+)
+
+type EnvBucket struct {
+	Items map[string]string
+}
+
+func NewEnvBucket() EnvBucket {
+	return EnvBucket{
+		Items: make(map[string]string),
+	}
+}
+
+func (e EnvBucket) Getenv(key string) string {
+	return e.Items[key]
+}
+
+func (e EnvBucket) Setenv(key string, value string) {
+	e.Items[key] = value
+}
+
+func TestRead_EmptyTimeoutConfig(t *testing.T) {
+	defaults := NewEnvBucket()
+	readConfig := ReadConfig{}
+
+	config := readConfig.Read(defaults)
+
+	if (config.readTimeout) != time.Duration(8)*time.Second {
+		t.Log("readTimeout incorrect")
+		t.Fail()
+	}
+	if (config.writeTimeout) != time.Duration(8)*time.Second {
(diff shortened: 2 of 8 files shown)


gateway/readconfig.go
```diff
@@ -0,0 +1,74 @@
...
+
+	cfg.readTimeout = time.Duration(readTimeout) * time.Second
+	cfg.writeTimeout = time.Duration(writeTimeout) * time.Second
+
+	return cfg
+}
+
+// GatewayConfig for the process.
+type GatewayConfig struct {
+	httpPort     int
```
alexellis: Let's not make the httpPort configurable - there are assumptions that the port is fixed.

gateway/server.go
```diff
@@ -20,6 +19,14 @@ func main() {
 	logger := logrus.Logger{}
 	logrus.SetFormatter(&logrus.TextFormatter{})
 
+	osEnv := OsEnv{}
+	readConfig := ReadConfig{}
+	config := readConfig.Read(osEnv)
+
+	log.Printf("HTTP Port: %d", config.httpPort)
```
alexellis: Please remove the HTTP port config.

gateway/readconfig.go
```diff
@@ -0,0 +1,74 @@
...
+// Read fetches config from environmental variables.
+func (ReadConfig) Read(hasEnv HasEnv) GatewayConfig {
+	cfg := GatewayConfig{
+		httpPort: 8080,
+	}
+
+	readTimeout := parseIntValue(hasEnv.Getenv("read_timeout"))
+	writeTimeout := parseIntValue(hasEnv.Getenv("write_timeout"))
+
+	if readTimeout == 0 {
```
alexellis: Let's modify parseIntValue to take a default override value as a parameter - therefore if parsing fails we get the default value instead. That will make the code a bit more concise too.

gateway/readconfig.go
stealthybox: This was still hard-coded -- just moved it into the config struct.
I have removed it since it was out of scope.

gateway/readconfig.go
stealthybox: Awesome suggestion -- implemented it and it's much cleaner.

stealthybox:
Thanks for the review -- changes are pushed.

gateway/config_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (e EnvBucket) Getenv(key string) string {
+	return e.Items[key]
+}
+
+func (e EnvBucket) Setenv(key string, value string) {
+	e.Items[key] = value
+}
+
+func TestRead_ReadAndWriteTimeoutConfig(t *testing.T) {
```
alexellis: We also need a unit test to show the default behaviour of 8 seconds for both variables if they are not set.

gateway/Dockerfile.build
```diff
@@ -10,13 +10,16 @@ FROM golang:1.7.5
...
-COPY requests   requests
-COPY tests      tests
-COPY server.go  .
+COPY handlers       handlers
+COPY metrics        metrics
+COPY requests       requests
+COPY tests          tests
+COPY server.go      .
+COPY readconfig.go  .
+COPY config_test.go .
```
alexellis: Would it be tricky to put config_test.go into the tests folder?

gateway/config_test.go
stealthybox: We should add one for the watchdog as well ( unless I missed it )

gateway/Dockerfile.build
stealthybox: I started with it in the tests folder when I had this realization ^

gateway/config_test.go
alexellis: "Go" for it.

stealthybox: Watchdog has a test for empty config -- I missed it the first time.
Added to gateway /w 8 sec.

gateway/Dockerfile.build
alexellis: Let's export them then.

stealthybox: @alexellis I pushed a commit that exports the fields and moves `config_test` to the *tests* package.

alexellis:
@stealthybox thanks for the updates. I've got this on my list for testing / merging.

stealthybox:
@alexellis appreciate the critique -- looking forward to it :whale:

stealthybox:
🎉

PR #98, opened 2017-07-02
Fix Watchdog Timeouts - duplicate multiplication of `time.Second`

### LINKED ISSUE #97: Watchdog Timeouts are multiplied by time.Second^2
## Expected Behaviour
Watchdog should timeout after 5 seconds (default) or the configured amount of seconds.

## Current Behaviour

Resulting timeout values are very large.
It's not possible to configure small values.
This also affects the default.

## Possible Solution

Remove the extra arithmetic in `main.go`.

## Context

This doesn't look intentional.
I'm not sure what the repercussions are.

I discovered this when copying the timeout configurations over to the Gateway httpServer.

stealthybox:
Removed duplicate multiplication of `time.Second` in *main.go*.
Resolves #97 

## Description
Times are already parsed as ints and multiplied by seconds before they are stored in the config struct as Durations.
They do not need further calculation when being passed to the httpServer, so the code has been removed.

## Motivation and Context

Functions should terminate within a timely fashion.
Currently, the timeout is much much much longer than expected.

#### sleeper function using `functions/alpine:health` with the default timeout of 5s:
```bash
λ curl localhost:8080/function/func_sleeper -XPOST -d7   # does not timeout
hi, sleeping for: 7
done
```

#### sleeper function using a fresh `functions/watchdog:build` with the deafult timeout:
```bash
λ curl localhost:8080/function/func_sleeper -XPOST -d2
hi, sleeping for: 2
done

λ curl localhost:8080/function/func_sleeper -XPOST -d5   # times out
Can't reach service: func_sleeper

λ curl localhost:8080/function/func_sleeper -XPOST -d4
hi, sleeping for: 4
done
```

## Types of changes

## Checklist:

## PS:
I've also added `--no-cache` to the dockerfile for when there is no `$http_proxy`.
This may or may not be desireable.

### DIFF
--- a/watchdog/build.sh
+++ b/watchdog/build.sh
@@ -7,7 +7,7 @@ if [ ! $http_proxy == "" ]
 then
     docker build --no-cache --build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy -t functions/watchdog:build .
 else
-    docker build -t functions/watchdog:build .
+    docker build --no-cache -t functions/watchdog:build .
 fi
 
 docker create --name buildoutput functions/watchdog:build echo
--- a/watchdog/main.go
+++ b/watchdog/main.go
@@ -147,13 +147,10 @@ func main() {
 		return
 	}
 
-	readTimeout := time.Duration(config.readTimeout) * time.Second
-	writeTimeout := time.Duration(config.writeTimeout) * time.Second
-
 	s := &http.Server{
 		Addr:           ":8080",
-		ReadTimeout:    readTimeout,
-		WriteTimeout:   writeTimeout,
+		ReadTimeout:    config.readTimeout,
+		WriteTimeout:   config.writeTimeout,
 		MaxHeaderBytes: 1 << 20,
 	}
 


watchdog/build.sh
```diff
@@ -7,7 +7,7 @@ if [ ! $http_proxy == "" ]
 then
     docker build --no-cache --build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy -t functions/watchdog:build .
 else
-    docker build -t functions/watchdog:build .
+    docker build --no-cache -t functions/watchdog:build .
```
alexellis: What was this change for?

watchdog/main.go
```diff
@@ -147,13 +147,10 @@ func main() {
 		return
 	}
 
-	readTimeout := time.Duration(config.readTimeout) * time.Second
```
alexellis: Good spot - this is actually duplicated now. But for testing you can do something like this:

```
$ cd faas/watchdog
$ go build
$ read_timeout=2 write_timeout=2 write_debug=true fprocess="sleep 8" ./watchdog
```

Then:

```
$ curl -v localhost:8080 -d "test"
```

alexellis:
Please test via #100 - which includes your change and also hard timeouts for the exec/forked process.

PR #100, opened 2017-07-03
Watchdog refurbishments

alexellis:
Watchdog refurbishments

## Description

- Watchdog - allow new methods with and without body.
- Enforce hard-timeout via exec_timeout variable - gives *408 Request Timeout*
- Correct bug in timeouts for read/write of HTTP. via @stealthybox #98
- Documentation for new verbs and hard timeout.
- cgi_headers now defaults to true for functions.

## How Has This Been Tested?

- Gateway unaffected by commits

- Tested through new and existing unit tests

- Tested through watchdog binary outside of container:

```
exec_timeout=2 fprocess="node test.js" ./watchdog
```

test.js:

```
function t() {
  console.log("Stop");
}
console.log("Start");

setTimeout(t, 4000);
```

Result:

```
 
$ curl localhost:8080 -v
* Rebuilt URL to: localhost:8080/
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.51.0
> Accept: */*
> 
< HTTP/1.1 408 Request Timeout
< Date: Mon, 03 Jul 2017 18:25:25 GMT
< Content-Length: 16
< Content-Type: text/plain; charset=utf-8
< 
Killed process.
* Curl_http_done: called premature == 0
* Connection #0 to host localhost left intact
```

## Types of changes

## Checklist:

### DIFF
--- a/watchdog/config_test.go
+++ b/watchdog/config_test.go
@@ -25,6 +25,32 @@ func (e EnvBucket) Getenv(key string) string {
 func (e EnvBucket) Setenv(key string, value string) {
 	e.Items[key] = value
 }
+
+func TestRead_CgiHeaders_OverideFalse(t *testing.T) {
+	defaults := NewEnvBucket()
+	readConfig := ReadConfig{}
+	defaults.Setenv("cgi_headers", "false")
+
+	config := readConfig.Read(defaults)
+
+	if config.cgiHeaders != false {
+		t.Logf("cgiHeaders should have been false (via env)")
+		t.Fail()
+	}
+}
+
+func TestRead_CgiHeaders_DefaultIsTrueConfig(t *testing.T) {
+	defaults := NewEnvBucket()
+	readConfig := ReadConfig{}
+
+	config := readConfig.Read(defaults)
+
+	if config.cgiHeaders != true {
+		t.Logf("cgiHeaders should have been true (unspecified)")
+		t.Fail()
+	}
+}
+
 func TestRead_WriteDebug_DefaultIsTrueConfig(t *testing.T) {
 	defaults := NewEnvBucket()
 	readConfig := ReadConfig{}
@@ -132,6 +158,20 @@ func TestRead_ReadAndWriteTimeoutConfig(t *testing.T) {
 	}
 	if (config.writeTimeout) != time.Duration(60)*time.Second {
 		t.Logf("writeTimeout incorrect, got: %d\n", config.writeTimeout)
+		t.Fail()
+	}
+}
+
+func TestRead_ExecTimeoutConfig(t *testing.T) {
+	defaults := NewEnvBucket()
+	defaults.Setenv("exec_timeout", "3")
+
+	readConfig := ReadConfig{}
+	config := readConfig.Read(defaults)
+
+	want := time.Duration(3) * time.Second
+	if (config.execTimeout) != want {
+		t.Logf("readTimeout incorrect, got: %d - want: %s\n", config.execTimeout, want)
 		t.Fail()
 	}
 }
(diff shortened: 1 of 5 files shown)


watchdog/main.go
```diff
@@ -115,24 +164,56 @@ func pipeRequest(config *WatchdogConfig, w http.ResponseWriter, r *http.Request)
...
-			pipeRequest(config, w, r)
-		} else {
+		switch r.Method {
+		case
+			"POST",
+			"PUT",
+			"DELETE",
+			"UPDATE":
+			pipeRequest(config, w, r, r.Method, true)
+			break
```
pmuens: Why is a `break` added here?

watchdog/main.go
```diff
@@ -115,24 +164,56 @@ func pipeRequest(config *WatchdogConfig, w http.ResponseWriter, r *http.Request)
...
+		case
+			"GET":
+			pipeRequest(config, w, r, r.Method, false)
+			break
```
pmuens: Same question for this `break`

watchdog/requesthandler_test.go
```diff
@@ -120,47 +121,102 @@ func TestHandler_HasXDurationSecondsHeader(t *testing.T) {
...
 	rr := httptest.NewRecorder()
 
 	req, err := http.NewRequest("GET", "/", nil)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	config := WatchdogConfig{}
+	config := WatchdogConfig{
+		// writeDebug:  true,
```
pmuens: Should this comment be removed?

alexellis:
Quick note on the GET method - `cat` will always fail for a `GET` - try changing it over to something that does not expect input such as `date` or `env`.

watchdog/config_test.go
```diff
@@ -25,6 +25,32 @@ func (e EnvBucket) Getenv(key string) string {
 func (e EnvBucket) Setenv(key string, value string) {
 	e.Items[key] = value
 }
+
+func TestRead_CgiHeaders_OverideFalse(t *testing.T) {
```
pmuens: `OverrideFalse`

watchdog/main.go
```diff
@@ -41,65 +44,111 @@ func debugHeaders(source *http.Header, direction string) {
...
+		}()
+	}
+
+	// Only write body if this is appropriate for the method.
+	if hasBody {
+		// Write to pipe in separate go-routine to prevent blocking
+		go func() {
+			defer wg.Done()
+			writer.Write(requestBody)
+			writer.Close()
```
pmuens: What about moving this to the top of the function and adding a `defer` here?

watchdog/main.go
alexellis: Golint was OK with it. It's more explicit about the flow.

watchdog/requesthandler_test.go
alexellis: Yes it can be

watchdog/main.go
alexellis: I think (should check) that this can make the close happen sooner.

alexellis:
I like those check marks. Thanks for your work and comments!

I would expect `cat` to block and then timeout. Is that what you saw?

watchdog/main.go
pmuens: Gotcha. 👍

alexellis:
> 🤔 just tried it today by running the compiled fwatchdog with fprocess="/bin/cat". Unfortunately no timeout when doing a GET request (waiting here for 2 minutes).

Try with the `exec_timeout` env-var set.

PR #106, opened 2017-07-12
Update vendoring via vndr

alexellis:
## Description

Housekeeping - update  to vendored dependencies - switched from `glide` to `vndr` for gateway due to issues resolving logrus.


alexellis:
Tested locally - scaling / prometheus etc working.

Issue #101, opened 2017-07-04
Proposal: functions should specify OS constraints


alexellis:
## Expected Behaviour

Deploying a function on a mixed-OS cluster should be possible - where a Windows container is scheduled correctly.

## Current Behaviour

The task will be scheduled randomly - possibly on a Linux host.

## Possible Solution

Add a default constraint to Linux and allow this to be overridden in the *deploy function* endpoint. A change to the CLI may also make sense later down the line.

Checkpoints:

* encompasses the RESTFul schema

Schema https://github.com/alexellis/faas/blob/master/gateway/requests/requests.go#L7

* Includes an update to the swagger API

* Covers both Kubernetes and Swarm

glyif:
I can try and take on this issue :)

alexellis:
I think @rgee0 started work on this.. is that still in-progress?

@glyif thanks very much for the offer of help, much appreciated. Can you get in touch with your email address for an invite to Slack?

glyif:
@alexellis my email is 122@holbertonschool.com :)

alexellis:
@rgee0 @glyif what's your bandwidth like for this week?

alexellis:
In lieu of this being completed #160 makes the default experience better by ensuring Linux containers are scheduled on Linux nodes.

alexellis:
Now available in the FaaS-CLI by passing --constraint or using a `constraints` element for a function. @glyif @rgee0 fancy writing a quick guide/blog?

PR #160, opened 2017-09-01
Linux placement by default

alexellis:
## Description

In a hybrid Swarm a newcomer to OpenFaaS has to learn about placement constraints very quickly... so let's set up some good defaults to make sure Linux containers are scheduled only on Linux nodes.

Anyone who wants to schedule a Windows function can do so through the docker-compose YAML file - or via #101 when complete. I believe @rgee0 has this in progress.

Swarm supports hybrid clusters of Windows 2016 + Linux.

Alex


docker-compose.extended.yml
```diff
@@ -5,16 +5,13 @@ services:
...
             - 8080:8080
-        image: functions/gateway:0.6.2 
+        image: functions/gateway:0.6.3 
         networks:
             - functions
         environment:
             dnsrr: "true"  # Temporarily use dnsrr in place of VIP while issue persists on PWD
             faas_nats_address: "nats"
             faas_nats_port: 4222
-        deploy:
```
alexellis: @developius we had `placement` for the gateway twice which is how you managed to schedule it on a 'worker' node.

alexellis:
Passes CI + tests - merging.

docker-compose.extended.yml
developius: Missed this. Makes sense

PR #166, opened 2017-09-05
[WIP] Use Prometheus API for function metrics - fixes #116

### LINKED ISSUE #116: API / UI - use Prometheus API (not local counter) for stats
## Expected Behaviour

Now that the Gateway can be load-balanced an invocation count should cover the count across all of the gateways.

## Current Behaviour

For Swarm it shows the current gateway instance since we're reading the local counter

For Kubernetes it shows as empty

## Possible Solution

Query the Prometheus API from the gateway:

PromQL example:

```
gateway_function_invocation_total{}
```

Concerns:

* creates a circular dependency since the gateway is a scrape target
* scrape interval is around 5-15s meaning we won't see the invocation count increase live.

* this data needs to be overlaid on top of the "list functions" response from Swarm or the external gateway - See: `faasHandlers.ListFunctions` in `server.go` for the FaaS gateway

## Your Environment

Both

johnmccabe:
**Work in progress** - Do not merge, raising for feedback

## Description

This commit adds a first run at gathering function invocation counts directly from Prometheus.

This is done in the handler returned from MakeFunctionReader().

- [ ] Error handling when unable to contact the Prometheus endpoint
- [ ] Handling for external provider - need to test with faas-netes (@alexellis worth making the Prometheus query endpoint configurable?)

## Motivation and Context

- [x] I have raised an issue to propose this change ([required](https://github.com/alexellis/faas/blob/master/CONTRIBUTING.md))

See #116 

## How Has This Been Tested?

Have built the gateway locally and deployed the stack with the new gateway image (updating the image in `docker-compose.yml` to `image: functions/gateway:latest-dev`).

Scale out the gateway to 3 nodes - `docker service scale func_gateway=3`

Opened the UI in two different browsers (Chrome and again in Incognito - see the comments in #116, it was possible to see the independent counts this way prior to this change) and invoked the `func_echoit` function multiple times from each browser, it was possible to observe the counts updating in both to the same value (based on the scrape interval in Prometheus).

## Types of changes

## Checklist:

### DIFF
--- a/gateway/handlers/reader.go
+++ b/gateway/handlers/reader.go
@@ -7,11 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"log"
 	"net/http"
-	"os"
-	"strconv"
 	"strings"
 
 	"github.com/alexellis/faas/gateway/metrics"
@@ -27,8 +23,6 @@ import (
 func MakeFunctionReader(metricsOptions metrics.MetricOptions, c *client.Client) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 
-		client := http.Client{}
-
 		serviceFilter := filters.NewArgs()
 
 		options := types.ServiceListOptions{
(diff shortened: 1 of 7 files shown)


gateway/handlers/reader.go
```diff
@@ -38,11 +43,33 @@ func MakeFunctionReader(metricsOptions metrics.MetricOptions, c *client.Client)
 		// TODO: Filter only "faas" functions (via metadata?)
 		var functions []requests.Function
 
+		req, reqErr := http.NewRequest("GET", "http://prometheus:9090/api/v1/query/?query=gateway_function_invocation_total", nil)
+		if reqErr != nil {
+			log.Println(reqErr)
```
johnmccabe: Ignore the use of log here, will strip that out

johnmccabe:
@alexellis is this what you were thinking.

alexellis:
It needs to be more generic so that it works one level above the handler.

So imagine you have figured out the handler we need for listing functions - it could be either an external handler (proxy) or the Swarm handler.

We should wrap that / put an adapter around it which calls the handler then massages the Prometheus data into it.

johnmccabe:
@alexellis this is where I was going with the enricher, could potentially hang more off this than just syncing metrics.

wdyt?

alexellis:
Thanks for working on this. I think with some tweaks it could be a useful pattern.

PR #171, opened 2017-09-07
Query Prometheus API for stats.

### LINKED ISSUE #116: API / UI - use Prometheus API (not local counter) for stats

alexellis:

## Description

Fixes #116 - related #166 

Also provides metrics for all external providers.

## Motivation and Context

- [x] I have raised an issue to propose this change ([required](https://github.com/alexellis/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?

Via Hyper.sh, Docker Swarm and Kubernetes

## Types of changes

## Checklist:

### DIFF
--- a/gateway/metrics/add_metrics.go
+++ b/gateway/metrics/add_metrics.go
@@ -1,0 +1,110 @@
+package metrics
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+
+	"github.com/alexellis/faas/gateway/requests"
+)
+
+func makeClient() http.Client {
+	// Fine-tune the client to fail fast.
+	return http.Client{}
+}
+
+// AddMetricsHandler wraps a http.HandlerFunc with Prometheus metrics
+func AddMetricsHandler(handler http.HandlerFunc, host string, port int) http.HandlerFunc {
+	client := makeClient()
+
+	prometheusQuery := NewPrometheusQuery(host, port, &client)
+	return func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


gateway/metrics/add_metrics.go
```diff
@@ -0,0 +1,110 @@
...
+			w.WriteHeader(http.StatusOK)
+			w.Write(upstreamBody)
+			return
+		}
+
+		mixIn(&functions, results)
+
+		bytesOut, marshalErr := json.Marshal(functions)
+		if marshalErr != nil {
+			log.Println(marshalErr)
```
johnmccabe:
```
w.WriteHeader(http.StatusInternalServerError)
w.Write([]byte("Error parsing metrics from upstream provider/backend."))
```

alexellis: At this point we have already parsed the two data sets and are now re-serializing them for transfer to the caller.

johnmccabe: Ah ok, I'd blackhole `_` the error in that case

Issue #175, opened 2017-09-09
Proposal: Support multiple different gateways with filtering options for function services


kinghuang:
This proposal relates to #55.

The OpenFaaS API Gateway discovers function services in a Docker Swarm based on the presence of a `function` label (or the `func` label on external providers). This behaviour is problematic if there are multiple instances of the API Gateway running, where each gateway is meant to cover different sets of services.

For example, in a multi-tenant environment, there may be more than one API Gateway deployed within the same swarm for different purposes. Or, there may be multiple deployments of a single stack for different branches of development. Functions meant for a certain API Gateway “leak” into other gateways because individual gateways do not have the ability to determine which specific functions are meant for them.

There are several possible ways to support multiple instances of the API Gateway in a single swarm.

### A: Allow the `function` label to be customized

`function` is currently coded as a literal string in the gateway to filter for OpenFaaS functions. This could be made into a configurable item so that gateways can be set to look for custom labels.

Here's an example of a gateway and a function configured to use `custom_function` as a label for function services.

```
gateway:
    image: functions/gateway:latest-dev
    environment:
        faas_function_label: 'custom_function'

echoit:
    image: functions/alpine:health
    labels:
        custom_function: "true"
    environment:
        fprocess: "cat"
```

#### Pros

* Simple to change. The string `function` only appears in 4 places. A config can be added for something like `faas_fuctions_label` that defaults to `function` and is passed where needed.
* Functions can have multiple labels to associate themselves with multiple distinct gateways.

#### Cons

* A changing label key (as opposed to the label value) might not be desirable.

### B: Look for a specific value or pattern in the `function` label

The value of the `function` label is not considered when reading for function services in `MakeFunctionReader()`. It's typically set to `"true"` in examples. A config could be added to specify a value or pattern to match on. So, something like the following (from [gateway/reader.go#L41](https://github.com/alexellis/faas/blob/master/gateway/handlers/reader.go#L41)):

Before:
```
if len(service.Spec.TaskTemplate.ContainerSpec.Labels["function"]) > 0 {
```

After:
```
label_value := service.Spec.TaskTemplate.ContainerSpec.Labels["function"]
match, _ := regexp.MatchString(label_value_pattern, label_value)

if match {
```

A similar change would apply in `swarmwatcher.AttachSwarmWatcher()`. Here's an example service definition for gateway and a function.

```
gateway:
    image: functions/gateway:latest-dev
    environment:
        faas_label_value: 'some-custom-value'

echoit:
[...]

Issue #55, opened 2017-04-19
Distinguish between functions belonging to different gateways


kinghuang:
## Expected Behavior

If there are multiple gateways in the same swarm, it should be possible to configure the gateways to pick up specific sets of functions, perhaps by labels. Otherwise, each gateway picks up all function services in the swarm.

## Current Behavior

Function services are currently denoted by the presence of a `function` label on the service. Gateways automatically pick up all services with this label.

## Possible Solution

As an example, [Traefik](https://traefik.io) can be [configured with constraints](https://docs.traefik.io/toml/#constraints) for which Traefik configured services it handles. Only services that match the constraint (via tags on labels) will be handled by the Traefik instance. This enables multiple instances of Traefik to be deployed on a swarm, each handling a different set of services.

## Context

I'm trying out FaaS in a project. The repo branches for this project are automatically built and deployed on a Docker swarm for review as part of a CI/CD process. The FaaS gateways from the deployments are picking up the functions from all the deployments, not just the specific branch. There doesn't seem to be a way to limit each gateway to the services for a specific deployment.

## Your Environment
* Docker version `docker version` (e.g. Docker 17.0.03 ): 17.0.04
* Operating System and version (e.g. Linux, Windows, MacOS): Development on macOS. CI/CD deployments on RHEL 7.3.
* Link to your project: Not available (internal).

alexellis:
Hi @kinghuang,

Thanks for opening this issue. I'd like to understand your use-case a little more.

In terms of two making FaaS gateways available - have you tried deploying individual stacks on separate networks? Can you share your YAML file with me?

```
# docker stack deploy pipeline1 -c ./stack1.yml
# docker stack deploy pipeline2 -c ./stack2.yml
```

alexellis:
I could reproduce the issue with the steps above. Both gateways saw the functions from either stack.

How would you propose injecting the unique network name / namespace into the gateway process/container?

alexellis:
You can use this test implementation - and provide a `stack_namespace` environmental variable to your gateways. This will filter your functions to only the gateway / stack they belong to.

https://github.com/alexellis/faas/compare/filter_networks?expand=1

alexellis:
@kinghuang please can you give feedback on above?

kinghuang:
@alexellis Will do! I've added this to my task list.

kinghuang:
[...]

PR #178, opened 2017-09-10
[WIP] Function label constraints

kinghuang:
A WIP implementation of Option E from #175 to allow service label filters to be configured on the gateway.

## Description

Add an environment config named `faas_function_filters` to the gateway. It is a comma-separated list of label filters used filter Docker services when reading the list of functions for the gateway. It is parsed into a string array and stored in a new field named `FunctionFilters` in `GatewayConfig`.

## Motivation and Context

When multiple gateways are deployed on a swarm, there is no way to configure gateways to pick up only certain functions. As a result, functions from different stacks or different deployments of the same stack “leak” into each other. This is a WIP implementation of one of the possible options for configuring the gateway to read only certain function services.

This implementation is for Option E in proposal issue #175, and aims to solve issue #55.

- [X] I have raised an issue to propose this change ([required](https://github.com/alexellis/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?

The `docker-compose.yml` file in this branch has been temporarily changed to filter for `hello=world`, and the `hubstats` function has been labeled to match. The gateway correctly picks up just that one function. If the `faas_function_filters` environment variable is removed, all the functions in the stack are read as before.

This currently breaks adding a new function in the UI, because the created service lacks the labels specified in `faas_function_filters`.

Also, this filtering is based on labels on services, not containers. It might be worth considering moving the `function` label to the service level for uniformity. However, that would be a breaking change.

## Types of changes

This becomes a breaking change if the `function` label is moved to the service level.

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I've read the [CONTRIBUTION](https://github.com/alexellis/faas/blob/master/CONTRIBUTING.md) guide

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,11 +5,12 @@ services:
             - "/var/run/docker.sock:/var/run/docker.sock"
         ports:
             - 8080:8080
-        image: functions/gateway:0.6.3
+        image: functions/gateway:latest-dev
         networks:
             - functions
         environment:
             dnsrr: "true"  # Temporarily use dnsrr in place of VIP while issue persists on PWD
+            faas_function_filters: "hello=world"
         deploy:
             placement:
                 constraints:
@@ -82,6 +83,8 @@ services:
             no_proxy: "gateway"
             https_proxy: $https_proxy
         deploy:
+            labels:
+                hello: "world"
             placement:
                 constraints:
                     - 'node.platform.os == linux'
--- a/gateway/handlers/reader.go
+++ b/gateway/handlers/reader.go
@@ -19,10 +19,13 @@ import (
 )
 
 // MakeFunctionReader gives a summary of Function structs with Docker service stats overlaid with Prometheus counters.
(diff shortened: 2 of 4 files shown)


alexellis:
Related: https://github.com/alexellis/faas-netes/issues/20

alexellis:
@kinghuang - where are you with this? Btw we have merge conflicts and commits need squashing.

docker-compose.yml
```diff
@@ -5,11 +5,12 @@ services:
             - "/var/run/docker.sock:/var/run/docker.sock"
         ports:
             - 8080:8080
-        image: functions/gateway:0.6.3
+        image: functions/gateway:latest-dev
```
alexellis: This line shouldn't have changed.

alexellis:
@kinghuang please see changes in latest release of the gateway and CLI to support labels.

alexellis:
Closing as inactive. I believe we have this supported in the code now via other PRs.

PR #188, opened 2017-09-14
Adding Kong guide

jonocodes:
Documentation only

### DIFF
--- a/guide/kong_integration.md
+++ b/guide/kong_integration.md
@@ -1,4 +1,4 @@
-# Securing OpenFaaS with Kong
+# Integrating OpenFaaS with Kong
 
 [Kong](https://getkong.org) is an API gateway that provides features such as security, logging, and rate limiting. By putting this in front of OpenFaaS you can quickly get access to these things and a lot more via [the many other plugins written](https://getkong.org/plugins/) for it.
 
@@ -25,6 +25,7 @@ hello world
 ```
 $ docker service create --name kong-database \
     --network func_functions --detach=false \
+    --constraint 'node.role == manager' \
     -e "POSTGRES_USER=kong" \
     -e "POSTGRES_DB=kong" \
     -e "POSTGRES_PASSWORD=secretpassword" \
@@ -39,6 +40,7 @@ $ docker service create --name=kong-migrations \
 
 $ docker service create --name kong \
     --network func_functions --detach=false \
+    --constraint 'node.role == manager' \
     -e "KONG_DATABASE=postgres" \
     -e "KONG_PG_HOST=kong-database" \
     -e "KONG_PG_PASSWORD=secretpassword" \
@@ -48,13 +50,16 @@ $ docker service create --name kong \
     -e "KONG_ADMIN_ERROR_LOG=/dev/stderr" \
     -p 8000:8000 \
     -p 8443:8443 \
-    -p 8001:8001 \
     kong:latest
 ```
 
-See that Kong us up and running
+Create a curl command alias so we can talk to the Kong admin without exposing its ports to the network.
 ```
-$ curl -i localhost:8001
+$ alias kong_admin_curl='docker exec $(docker ps -q -f name="kong\.") curl'
+```
+See that Kong admin is up and running
+```
+$ kong_admin_curl -i localhost:8001
 HTTP/1.1 200
 ...
 ```
@@ -63,7 +68,7 @@ HTTP/1.1 200
 
 Proxy OpenFaaS's functions through Kong
 ```
-$ curl -i -X POST \
+$ kong_admin_curl -X POST \
     --url http://localhost:8001/apis/ \
     --data 'name=function' \
     --data 'uris=/function' \
@@ -73,15 +78,12 @@ $ curl localhost:8000/function/func_echoit -d 'hello world'
 hello world
 ```
 
-In order to benefit from the security Kong gives you, you should make sure only to expose Kong's public port (in this case its 8000) through your firewall. If you keep 8080 exposed, then the security Kong gives you can be circumvented.
-
-
 ### Require basic authentication
 
 Enable the basic-auth plugin in Kong
 
 ```
-$ curl -X POST http://localhost:8001/plugins \
+$ kong_admin_curl -X POST http://localhost:8001/plugins \
     --data "name=basic-auth" \
     --data "config.hide_credentials=true"
 ```
@@ -89,9 +91,9 @@ $ curl -X POST http://localhost:8001/plugins \
 Create a consumer with credentials
 
 ```
-$ curl -d "username=aladdin" http://localhost:8001/consumers/
+$ kong_admin_curl -d "username=aladdin" http://localhost:8001/consumers/
 
-$ curl -X POST http://localhost:8001/consumers/aladdin/basic-auth \
+$ kong_admin_curl -X POST http://localhost:8001/consumers/aladdin/basic-auth \
     --data "username=aladdin" \
     --data "password=OpenSesame"
 ```
@@ -117,7 +119,7 @@ hello world
 Now lets expose the /ui directory so we can securely use the web GUI
 
 ```
-$ curl -i -X POST \
+$ kong_admin_curl -i -X POST \
     --url http://localhost:8001/apis/ \
     --data 'name=ui' \
     --data 'uris=/ui' \
@@ -127,7 +129,7 @@ $ curl -i -X POST \
 Additionally we need to expose /system/functions since the UI makes Ajax requests to it
 
 ```
-$ curl -i -X POST \
+$ kong_admin_curl -i -X POST \
     --url http://localhost:8001/apis/ \
     --data 'name=system-functions' \
     --data 'uris=/system/functions' \
@@ -161,19 +163,19 @@ $ openssl req -x509 -nodes -days 365 -newkey rsa:2048 \
 Add cert to Kong
 
 ```
-$ curl -i -X POST http://localhost:8001/certificates \
-    -F "cert=@/tmp/selfsigned.pem" \
-    -F "key=@/tmp/selfsigned.key" \
+$ kong_admin_curl -X POST http://localhost:8001/certificates \
+    -F "cert=$(cat /tmp/selfsigned.pem)" \
+    -F "key=$(cat /tmp/selfsigned.key)" \
     -F "snis=example.com"
 
 HTTP/1.1 201 Created
 ...
 ```
 
-Use the cert to secure OpenFaaS
+Put the cert in front OpenFaaS
 
 ```
-$ curl -i -X POST http://localhost:8001/apis \
+$ kong_admin_curl -i -X POST http://localhost:8001/apis \
     -d "name=ssl-api" \
     -d "upstream_url=http://gateway:8080" \
     -d "hosts=example.com"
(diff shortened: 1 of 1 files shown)


ems5311:
Trying this out, I'm getting the following error when attempting to access my function(s) through the Kong API:

```
$ curl -i -X POST   --url http://localhost:8001/apis/   --data 'name=echoit'   --data 'uris=/echo'   --data 'upstream_url=http://localhost:8080/function/func_echoit'
$ curl localhost:8000/echo -d 'hello there'
curl: (6) Could not resolve host: GET
HTTP/1.1 502 Bad Gateway
Date: Fri, 15 Sep 2017 15:56:54 GMT
Content-Type: text/plain; charset=UTF-8
Transfer-Encoding: chunked
Connection: keep-alive
Server: kong/0.11.0

An invalid response was received from the upstream server

When I do a search engine query for this error, I find this post which may be helpful: https://github.com/Mashape/kong/issues/163#issuecomment-152634228

Do we need to have the API running in the same kong container?

**Comment by jonocodes** [comment]
Yeah @ems5311, that wont work. At least not in the example, or in a standard docker setup.

You cant use localhost in the upstream url because from Kong's perspective it is running on localhost, not FaaS. FaaS is running on a different node in a docker swarm. Note my use of the $FAASHOST variable which is my way of getting a route that Kong can use.

You would be able to use localhost if you did the standard FaaS setup, and then installed Kong right on the local machine instead of running it in a container however. But I did not do that in the sample since setup is more difficult.

**Comment by alexellis** [comment]
How about an entry in the hosts file?

**Comment by alexellis** [comment]
We have to get the dco

**Comment by jonocodes** [comment]
Not exactly sure what the dco tag is about but if its about signoff, I did correct that with a later commit. Let me know if I should do that commit in a cleaner way perhaps.

In terms of the hosts file... sure you can do that. Basically we could modify the docker run command for kong to have something like this in it:
`docker run --add-host=faashost:$FAASHOST .... kong:latest`

and then you can set the upstream_url host to faashost.

But as you can see this is not really any cleaner then how it is already. I guess it depends on what you goal is here. Are you trying to simplify the stack?

I guess it depends on what is trying to be shown in this doc. 
Is it a sample of how to integrate services, that people can modify for their own purposes/topology? That is how it is written how.

Or is it a quick way to inject authentication into your running production FaaS? Then perhaps it would be better to get Kong running in the same swarm/network as FaaS is already using.

**Comment by alexellis** [comment]
You'll need to sign all commits in the PR - the easiest way to do that is to squash them into one then repush with `--force`.

**Comment by alexellis** [comment]
Kong should ideally also be offered via `docker service create` rather than run - probably inside the same network which is `func_functions` i.e. the stack name plus underscore plus the network name.

**Comment by ems5311** [comment]
@jonocodes Thanks for your help. I still get the error `502 Bad Gateway` error when setting the upstream_url correctly to `$(hostname -i)`. I also tried adding the host with `--add-host=faashost:$FAASHOST` in the `docker run -d --name kong ...` line, but it still returns a 502 Bad Gateway error. I made sure that both the kong container and kong db container are up and running, I'm going to see if I can find some logs for kong.

**Comment by jonocodes** [comment]
Ok. I have cleaned up the commits and brought up Kong as a service in the same network as FaaS.

**Comment by ems5311** [comment]
Works for me now, :100: . Thanks!!

File: guide/kong_integration.md [hunk 1]
```diff
@@ -0,0 +1,96 @@
+# Using Kong as an API gateway for OpenFaaS
+
+[Kong](https://getkong.org) is an API gateway that provides features such as security, logging, and rate limiting. By putting this in front of OpenFaaS you can quickly get access to these things and a lot more via [the many other plugins written](https://getkong.org/plugins/) for it.
+
+Below is a demo of how you could use Kong as an authentication layer for OpenFaaS. You should be able to paste this all (from its Markdown source) into [Play With Docker](http://labs.play-with-docker.com/) to see it in action.
```
alexellis: Should?

guide/kong_integration.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Setup OpenFaaS
+
+    docker swarm init --advertise-addr $(hostname -i)
```
alexellis: General established documentation format is to use triple block-quotes ``` and to prefix input with `$` and output without any prefix.

The services look good :+1: I think there is also a chance we could use a Docker Stack here.

guide/kong_integration.md
```diff
@@ -0,0 +1,96 @@
...
+      --url http://localhost:8001/apis/ \
+      --data 'name=functions' \
+      --data 'uris=/functs' \
+      --data 'upstream_url=http://gateway:8080/function'
+
+    curl localhost:8000/functs/func_echoit -d 'hello there'
+
+
+## Add a some auth with a Kong plugin
+
```
alexellis: Good clear steps. Can we add a note on what each line is doing?

guide/kong_integration.md
jonocodes: I can change 'should be able to' to 'can'.

guide/kong_integration.md
jonocodes: I can change to block quotes and input and outputs. But my intent as stated above is that you can cut and paste this whole from markdown source into play-with-docker or your local terminal to test it.

Thats how I built and and thats how I test it. Since headings are prefixed with hash marks they are interpreted as comments. If I put in block quotes each section will have to be separately pasted by the user. If we start each command with $, then even more cutting and pasting would be needed.

Actually what would be really nice would be implementing a PWD lab with clickable text sections so the user doesnt even have to cut and paste - but then again you can see I'm a little over excited about PWD :-)

I guess I favor keeping this doc easily testable, but I if someone feels this diverges too much from an established documentation format, I'm OK with changing it.

alexellis: I know blocks can't be copied and pasted but it's the established style and slows the tutorial down too. This is in-line with the Docker and Arch Linux documentation.

You can submit a PWD lab for this, I wrote one for multi-stage builds that you can use as a template.

jonocodes: Concerning the docker stack... are you talking about modifying the faas/docker-compose.yml file to include kong? Or maybe creating something like docker-coompose.secure.yml that will do this?

alexellis: I can imagine this being a core use-case for people. In particular Kubernetes also uses a "stacks" concept through YAML files. We should test this out against FaaS-netes and provide a "YAML" configuration similar to: https://github.com/alexellis/faas-netes/blob/master/monitoring.yml

alexellis:
After we update the style I'll merge the guide. Talked off thread with @jonocodes about writing additional steps for TLS termination and basic-auth for the UI routes.

guide/kong_integration.md
jonocodes: Hm. Perhaps a separate tight for that one.

coopr:
Bravo!

jonocodes:
I have updated it to be a more general security guide using Kong.

guide/kong_integration.md
```diff
@@ -0,0 +1,196 @@
...
+```
+$ curl localhost:8080/function/func_echoit -d 'hello world'
+hello world
+```
+
+## Setup Kong
+```
+$ docker service create --network func_functions --detach=false \
+      --name kong-database \
+      -p 5432:5432 \
```
johnmccabe: I don't think you need to expose this, both the `kong` and `kong-migrations` containers can access this without the mapping over the `func_functions` network.

guide/kong_integration.md
```diff
@@ -0,0 +1,196 @@
...
+$ docker service create --network func_functions --name kong \
+      -e "KONG_DATABASE=postgres" \
+      -e "KONG_PG_HOST=kong-database" \
+      -e "KONG_PROXY_ACCESS_LOG=/dev/stdout" \
+      -e "KONG_ADMIN_ACCESS_LOG=/dev/stdout" \
+      -e "KONG_PROXY_ERROR_LOG=/dev/stderr" \
+      -e "KONG_ADMIN_ERROR_LOG=/dev/stderr" \
+      -p 8000:8000 \
+      -p 8443:8443 \
+      -p 8001:8001 \
```
johnmccabe: Worth adding a short firewall section to highlight all the admin/api ports that should be restricted in one place perhaps?

guide/kong_integration.md
```diff
@@ -0,0 +1,196 @@
...
+      -e "POSTGRES_DB=kong" \
+      postgres:9.4
+
+$ docker service create --network func_functions --detach=false \
+      --restart-condition=none --name=kong-migrations \
+      -e "KONG_DATABASE=postgres" \
+      -e "KONG_PG_HOST=kong-database" \
+      kong:latest kong migrations up
+
+$ docker service create --network func_functions --name kong \
```
johnmccabe: `--detach=false` missing?

guide/kong_integration.md
```diff
@@ -0,0 +1,196 @@
...
+hello world
+```
+
+## Setup Kong
+```
+$ docker service create --network func_functions --detach=false \
+      --name kong-database \
+      -p 5432:5432 \
+      -e "POSTGRES_USER=kong" \
+      -e "POSTGRES_DB=kong" \
```
johnmccabe: Set a password to secure the DB?

alexellis:
I love that this thread has 30 (now 31) comments already.

guide/kong_integration.md
alexellis: Is it possible to run with fewer ports exposed?

guide/kong_integration.md
alexellis: The network I believe is allowing us to specify the upstream by DNS name inside the overlay network such as `http://gateway:8080/`. Or what were you thinking?

alexellis:
@coopr Do your team have anything to add on this?

guide/kong_integration.md
johnmccabe: was that question for me?

alexellis: Yes - didn't understand your comment re: network mapping

johnmccabe: ah ok, yes the migration and main kong services are accessing the database over the `func_functions` network via the dns name. There is no need to expose the port as its only being used internally.

guide/kong_integration.md
jonocodes: It is possible. Of the bat I can think of 2 ways of doing this.
Lets suppose we only want to expose 8443 and not 8000 (kong http user port), or 80001 (kong admin port)
1. on the host machine, get the IP of kong and use that to make all the admin requests to kong.
2. instead of making the admin requests from the host, we can docker exec into the kong machine and do all the updates from there.

However, I do think this makes the readme more confusing and seems to diverge a bit from how kong is documented. If the user is really going to put this in prod, they should have some sort of firewall regardless, no? You dont want to expose your entire machine to the network.

Also, if you are concerned with exposing ports, we should also hide the OpenFaaS ports as well since only Kong really needs to know them.

guide/kong_integration.md
jonocodes: sure. I will not expose 5432. its not needed.

guide/kong_integration.md
jonocodes: sure

johnmccabe:
@jonocodes its a cracking guide, I found following it to be super easy, thanks for spending the time on it!!

guide/kong_integration.md
```diff
@@ -0,0 +1,195 @@
...
+$ curl -k https://localhost:8443/function/func_echoit \
+  -d 'hello world' -H 'Authorization: Basic YWxhZGRpbjpPcGVuU2VzYW1l'
+hello world
+```
+
+## Configure your firewall
+
+Between OpenFaaS and Kong a lot of ports are exposed on your host machine. In the end it is best to only expose either 8000 or 8443 out of your network depending if you added SSL or not.
+
+Another option is to expose both and enable [https_only](https://getkong.org/docs/0.11.x/proxy/#the-https_only-property) which is used to notify clients to upgrade to https from http.
```
coopr: If you change this URL to https://getkong.org/docs/latest/proxy/#the-https_only-property it'll automatically point to the latest release of Kong CE.

coopr:
Made one minor nice-to-have suggestion, otherwise LGTM, :shipit: ! I'll have some colleagues look at it soon, but don't wait for us to publish this - and thanks!!

alexellis:
I think we're pretty much there now. There is just one concern I have which is when we're using a cloud provider which doesn't use a firewall. I.e. Packet.

If the admin port 8001 is exposed then it will be accessible against any node in the cluster.

One way around this is to run Kong administration commands from within a Docker container (service) using `docker exec <container_id>`.

coopr:
We list other options for securing Kong's Admin API on https://getkong.org/docs/latest/secure-admin-api/

jonocodes:
Interesting @coopr. I do like reducing the footprint so I started looking at restricting the IP to the docker host, but that might get hairy depending on how their engine is setup.

The API loopback is also a good option, but would require me to either exec into the kong container anyway (suggested above). Or it would require me to disable the external 8001 port (perhaps via `docker update service --publish-rm 8001:8001 kong`), but I mentioned this to @alexellis and he did not like the idea of turning off open ports.

alexellis:
In the context of the admin port what I meant was - I'd rather never publish it than have to run ad-hoc commands afterwards to un-publish ports. Let's merge this guide and then think about a guide with a name like *Hardening OpenFaaS* where we can elaborate on this more.

PR #190, opened 2017-09-15
Added 'Integrating OpenFaaS and GraphQL'

kenfdev:
Added a blog post. @alexellis I also changed the author of my post but was that all right? Thanks!


alexellis:
Sorry it looks like I did that unintentionally. Thanks for fixing

PR #194, opened 2017-09-18
Update HACK.md

nullseed:
When creating the stack using an external network, this must be stated in the `docker-compose.yml` file otherwise the following error is reported:

```
alexs-macbook-pro:faas alex$ ./deploy_stack.sh
Deploying stack
Creating network func_functions
Error response from daemon: network with name func_functions already exists
```

## Motivation and Context

- [ ] I have raised an issue to propose this change ([required](https://github.com/alexellis/faas/blob/master/CONTRIBUTING.md))

## Types of changes

## Checklist:

### DIFF
--- a/contrib/HACK.md
+++ b/contrib/HACK.md
@@ -9,7 +9,7 @@ $ ./build.sh
 
 Now edit the gateway service in your `docker-compose.yml` file and deploy the stack.
 
-If you want to use an overriden name then pass in the tag to the `./build.sh` script such as `./build.sh test-1`.
+If you want to use an overridden name then pass in the tag to the `./build.sh` script such as `./build.sh test-1`.
 
 ## Hack on the UI for the API Gateway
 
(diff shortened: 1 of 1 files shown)


contrib/HACK.md
```diff
@@ -22,14 +22,20 @@ $ docker stack rm func
 $ docker network create func_functions --driver=overlay --attachable=true
 ```
 
-Now edit the `docker-compose.yml` file and add an attribute to the `functions` network of `attachable`.
+Now edit the `docker-compose.yml` file and replace the existing networks block with:
+
+```
+networks:
+    functions:
```
alexellis: So "attachable" isn't needed anymore?

nullseed: Not in the docker-compose.yml file because the network has been created with `docker network create func_functions --driver=overlay --attachable=true`

alexellis: Yes that's right.. I missed the line above in the diff. Cool let's merge it.

contrib/HACK.md
```diff
@@ -22,14 +22,20 @@ $ docker stack rm func
...
+            name: func_functions
+```
 
 Now you can run the gateway as its own container and bind-mount in the HTML assets.
 
 ```
-$ cd faas
 $ docker run -v `pwd`/gateway/assets:/root/assets -v "/var/run/docker.sock:/var/run/docker.sock" \
--p 8080:8080 --network=func_functions -ti functions/gateway:latest-dev
+-p 8080:8080 --network=func_functions -itd functions/gateway:latest-dev
```
alexellis: What's "-itd" vs "-it" ?

nullseed: The majority of the time when working with faas i've run the gateway container in daemon mode so i thought i'd be nice for those copying and pasting these commands to run the gateway in the background

alexellis: Should it be "-d" though? I've never seen "-itd"?

nullseed: They can be chained `-i -t -d` is equivalent to `-itd` https://docs.docker.com/engine/reference/commandline/container_run/#options

nullseed: Oh sorry you mean, we don't need the `-it` flags, yes you're right, it should be just `-d`

alexellis:
Merging. Thank you.

PR #292, opened 2017-10-13
[Testing] Use encrypted secrets

### LINKED ISSUE #285: Proposal: Secure secret management
This proposal is inspired by the Docker Swarm secret management but should be generalizable to Kubernetes Secrets and possibly other 3rd party secret management systems.

## Summary 
It is common that a function may need to connect to a database or a secure API. The username and password values used to connect to these data sources should be kept secret and secured.  Currently these values are either hardcoded into functions or provided as environment variable. This means that these values are not encrypted at rest and even more specifically it leads to the highly likely situation where these secret values are checked into git repositories. This puts these sensitive values at risk.

Both [Docker Swarm](https://docs.docker.com/engine/swarm/secrets/) and [Kubernetes](https://kubernetes.io/docs/concepts/configuration/secret/) have built in secret management systems.  In particular, both provide an API for management of secrets and support mounting those secrets in services as files.  Kubernetes is the most flexible in allowing you to specify the exact path of where these files are mounted.  Docker Swarm on the other hand strictly mounts the secrets to `/run/secrets` inside the containers.

## Expected behavior
OpenFaaS should support defining services that access secret values from and encrypted storage. 

## Current Behavior
Secrets can be provided via the Environment variables, but these values are not encrypted at rest.

## Possible solution
The end user should be able to provide secrets to the orchestration layer (docker swarm or kubernetes) and then reference those same secrets in the function. The creator of a function would simply provide a list of secrets that are required for the function during the function creation (or in the stack yaml for the cli) and the orchestration layer would be responsible for mounting those secrets as files in a standard location.

Specifically, 

1. We should standardize functions to expect secret files mounted in `/run/secrets`.
2. Secret values will be created and managed independently from functions via the orchestration tool
3. When creating a new function service, the API accepts a list of secret names as strings. The gateway then passes this forward to the orchestration layer when defining the new function service/pod.

```go
// CreateFunctionRequest create a function in the swarm.
type CreateFunctionRequest struct {
    // Service corresponds to a Docker Service
    Service string `json:"service"`

    // Image corresponds to a Docker image
    Image string `json:"image"`

    // Network is specific to Docker Swarm - default overlay network is: func_functions
    Network string `json:"network"`

    // EnvProcess corresponds to the fprocess variable for your container watchdog.
    EnvProcess string `json:"envProcess"`

    // EnvVars provides overrides for functions.
    EnvVars map[string]string `json:"envVars"`

    // Secrets is a list of secrets required for the orchestration layer to provide
    Secrets []string `json:"secrets"`

    // RegistryAuth is the registry authentication (optional)
    // in the same encoded format as Docker native credentials
    // (see ~/.docker/config.json)
    RegistryAuth string `json:"registryAuth,omitempty"`

    // Constraints are specific to back-end orchestration platform
    Constraints []string `json:"constraints"`
}
```

## Context

I want to use this functionality specifically to securely store database and api access credentials.

LucasRoesler:
## Description
- During function creation, accept an array of strings defining swarm secrets
that are required for the service
- Update docs
- Add new guide on using the secrets capability
- Add new sample function to highlight using environment variables
- Update `ApiKeyProtected` sample function to utilize the new secrets
capabilities

## Motivation and Context
This allows secrets to remain encrypted at rest instead of being unencrypted
in environment variables and yaml files.

Fixes #285

## How Has This Been Tested?
- I have validated that this does not break any of the integration tests
- Additionally I have tested it by updating the `ApiKeyProtected` function to use the secrets functionality 

## Types of changes

## Checklist:


LucasRoesler:
I just realized that we should support the more verbose secret specification like the docker cli does, see https://docs.docker.com/engine/reference/commandline/service_create/#create-a-service-with-secrets

If we do this, it allows 3rd parties to provide more sophisticated secrets handling, such as mount secret `foo` as file `bar`.   The docker docs give the following example

```sh
$ docker service create --name redis \
    --secret source=ssh-key,target=ssh
```

The docker cli implements this advanced parsing here: https://github.com/docker/cli/blob/1401d5daf2f49a97791487dd5c5a8598907f0bf1/opts/secret.go#L19

I am currently in the process of trying to vendor the `docker/cli/opts` but it is proving non-trivial.

alexellis:
Is it still passed as a list of strings?

guide/secure_secret_management.md
```diff
@@ -0,0 +1,75 @@
...
+
+OpenFaaS deploys functions as Docker Swarm Services, as result there are several features that we can leverage to simplify the development and subsquent deployment of functions to hardened production environments.
+
+First an most simple is the ability to set environment variables at deploy time. For example, you might want to set the `NODE_ENV` or `DEBUG` variable.  If you are interacting with the OpenFaaS gateway via the api, seeting the `NODE_ENV` might look like this
+
+
+```sh
+curl -H "Content-Type: application/json" \
+     -X POST \
+     -d '{"service":"nodeinfo","network":"func", "image": "functions/nodehelloenv:latest", "envVars": {"NODE_ENV": "production"}}' \
```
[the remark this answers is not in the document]
alexellis: Also might be worth showing this via the CLI not curl.. it's too low-level.

guide/secure_secret_management.md
```diff
@@ -0,0 +1,75 @@
...
+echo "R^Y$qzKzSJw51K9zP$pQ3R3N" | docker secret create secret_api_key -
+```
+
+Now, with the secret defined, we can deploy the function like this
+
+
+```sh
+curl -H "Content-Type: application/json" \
+     -X POST \
+     -d '{"service":"protectedapi","network":"func_functions", "image": "functions/api-key-protected:latest", "secrets": ["secret_api_key"]}' \
```
alexellis: Good to see re-use of the protected API example.

alexellis:
OK @LucasRoesler - marking as "WIP"

guide/secure_secret_management.md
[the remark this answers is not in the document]
LucasRoesler: I can switch it to the cli, but this example is supposed to demonstrate using the environment variables with the later example showing the usage of secrets.

LucasRoesler:
I just pushed a change that vendors the `docker/cli/opts`.  This allows more advanced specification of secrets.  Secrets are still provided as an array of strings, but those strings can now be a csv.  The update guide documents and example of using this functionality to control the name of the mounted secret.   Additionally, I updated the guide to show using secrets with the `faas-cli`.

I think this is ready now.

johnmccabe:
@alexellis I'll have a look at this when I get out of work, looks good at first very cursory glance

Something feels a bit iffy about pulling in the cli dependency for what is effectively a very light wrapper around the `github.com/docker/docker/api/types/swarm` package, we end up dragging in a lot of unused cruft there? I'm not sure that theres a better option tho..

gateway/vendor.conf
```diff
@@ -1,3 +1,4 @@
+github.com/docker/cli/opts 08097edc7849e1543d723a3c382536909b58aa4f
```
johnmccabe: this should be a root import not the package

LucasRoesler:
@johnmccabe  I agree, but I didn't see a good alternative other than simple copy and paste of https://github.com/docker/cli/blob/master/opts/secret.go#L19 and I didn't want to take credit for their work.

guide/secure_secret_management.md
```diff
@@ -0,0 +1,165 @@
...
+## Using Environment Variables
+First, and most simple, is the ability to set environment variables at deploy time. For example, you might want to set the `NODE_ENV` or `DEBUG` variable.  Setting the `NODE_ENV` while using the `faas-cli` might look like this
+
+```sh
+$ faas-cli deploy -f ./samples.yml
+$ fass-cli invoke nodehelloenv
+Hello from a production machine
+```
+
+Where your `samples.yml` stack file looks like this
```
johnmccabe: I'd consider moving the yaml before its invocation, as it flows more logically - ie here a secret in yaml, invoke the function, see the secret has been picked up

guide/secure_secret_management.md
```diff
@@ -0,0 +1,165 @@
...
+  nodehelloenv:
+    lang: Dockerfile
+    skip_build: true
+    image: functions/nodehelloenv:latest
+    environment:
+        NODE_ENV: production
+```
+
+
+(_Optional_) If you are directly using the OpenFaaS Gateway API, then it will look like this
```
johnmccabe: I had thought we were pushing people towards the CLI rather than providing curl examples - @alexellis wdyt?

guide/secure_secret_management.md
```diff
@@ -0,0 +1,165 @@
+# Guide on using Docker Swarm Secrets with OpenFaaS
+
+OpenFaaS deploys functions as Docker Swarm Services, as result there are several features that we can leverage to simplify the development and subsquent deployment of functions to hardened production environments.
+
+## Using Environment Variables
+First, and most simple, is the ability to set environment variables at deploy time. For example, you might want to set the `NODE_ENV` or `DEBUG` variable.  Setting the `NODE_ENV` while using the `faas-cli` might look like this
```
johnmccabe: Perhaps, "First, and least secure,"

guide/secure_secret_management.md
```diff
@@ -0,0 +1,165 @@
...
+$ curl -X POST \
+       -H 'Content-Type: text/plain' \
+       -H 'Content-Length: 0' \
+       http://localhost:8080/function/nodehelloenv
+Hello from a production machine
+```
+
+
+## Using Swarm Secrets
+A very tempting thing to do is to now add database password or api secrets as environment variables.  However, this is not secure.  Instead, we can leverage the [Docker Swarm Secrets](https://docs.docker.com/engine/swarm/secrets/) feature to safely store and give our functions access to the needed values. Using secrets is a two step process.  Take the [ApiKeyProtected](../sample-functions/ApiKeyProtected) example function, when we deploy this function we provide a secret key that it uses to authenticate requests to it.  First we must add a secret to the swarm
```
johnmccabe: If the envvar section begins pointing out its insecurity, I think you can drop the first two sentences here.

sample-functions/ApiKeyProtected/Dockerfile
```diff
@@ -1,3 +1,11 @@
+FROM golang:1.7.5 as builder
```
johnmccabe: giphy

sample-functions/ApiKeyProtected/README.md
```diff
@@ -1,6 +1,6 @@
 ### Api-Key-Protected sample
 
-To use this sample provide an env variable for the container/service in `secret_api_key`.
+To use this sample provide a secret for the container/service in `secret_api_key` using [Docker Swarm Secret](https://docs.docker.com/engine/swarm/secrets/#defining-and-using-secrets-in-compose-files).
```
johnmccabe: Might read better with "in the `secret_api_key` [Docker Swarm Secret].."

I don't quite follow the docker-compose link here, is this something to cover in the guide/needed?

gateway/handlers/createhandler.go
```diff
@@ -66,7 +75,8 @@ func MakeNewFunctionHandler(metricsOptions metrics.MetricOptions, c *client.Clie
 	}
 }
 
-func makeSpec(request *requests.CreateFunctionRequest, maxRestarts uint64, restartDelay time.Duration) swarm.ServiceSpec {
+func makeSpec(c *client.Client, request *requests.CreateFunctionRequest, maxRestarts uint64, restartDelay time.Duration) (swarm.ServiceSpec, error) {
```
johnmccabe: Perhaps move the secrets creation out of the spec creation and just pass it in, the spec creation is just an assembly. Makes it easier to test.

gateway/handlers/createhandler.go
```diff
@@ -66,7 +75,8 @@ func MakeNewFunctionHandler(metricsOptions metrics.MetricOptions, c *client.Clie
...
+	linuxOnlyConstraints := []string{"node.platform.os == linux"}
```
johnmccabe: this is already declared outside the function body

api-docs/swagger.yml
```diff
@@ -204,6 +204,10 @@ definitions:
         items:
           type: string
         description: Overrides to environmental variables
+      secrets:
+        type: array
+        items:
+          type: string
```
johnmccabe: Can you add a description

guide/secure_secret_management.md
```diff
@@ -0,0 +1,165 @@
...
+First add a new secret key
+
+```sh
+echo "new$qzKzSJw51K9zP$pQ3R3N" | docker secret create secret_api_key_2 -
+```
+
+Then, remove our old function and redeploy it with the new secret mounted in the same place as the old secret
+
+```sh
+$ echo "new$qzKzSJw51K9zP$pQ3R3N" | docker secret create secret_api_key_s -
```
johnmccabe: `_s` typo?

johnmccabe: this should be removed?

guide/secure_secret_management.md
```diff
@@ -0,0 +1,165 @@
...
+
+Unlocked the function!
+
+$ curl -H "Content-Type: application/json" \
+     -X POST \
+     -H "X-Api-Key: wrong_key" \
+     -d '{}' \
+     http://localhost:8080/function/protectedapi
+
+Access denied!
```
johnmccabe: Something a bit unexpected with the ApiKeyProtected example, its letting a sizeable substring match the key.
```
$ curl -H "Content-Type: application/json" -X POST -H "X-Api-Key: R^Y$qzKzSJw51K9zP$pQ3R3N" -d '{}' http://localhost:8080/function/protectedapi
Unlocked the function!

$ curl -H "Content-Type: application/json" -X POST -H "X-Api-Key: R^Y$q" -d '{}' http://localhost:8080/function/protectedapi
Unlocked the function!

$ curl -H "Content-Type: application/json" -X POST -H "X-Api-Key: R^Y$" -d '{}' http://localhost:8080/function/protectedapi
Access denied!

$ curl -H "Content-Type: application/json" -X POST -H "X-Api-Key: blahblahblah" -d '{}' http://localhost:8080/function/protectedapi
Access denied!
```
Are you seeing the same?

johnmccabe:
Looking pretty good, I've not dug into the secrets management on Swarm/K8s in detail yet so my comments are likely pretty superficial

@alexellis for the ApiKeyProtected example do we expect access to such an API to be managed by something like Kong with it injecting the `Http_X_Api_Key` header?

johnmccabe:
Looking pretty good, I've not dug into the secrets management on Swarm/K8s in detail yet so my comments are likely pretty superficial

@alexellis for the ApiKeyProtected example do we expect access to such an API to be managed by something like Kong with it injecting the `Http_X_Api_Key` header?

sample-functions/ApiKeyProtected/README.md
LucasRoesler: I will change it to link to the guide instead.

gateway/handlers/createhandler.go
LucasRoesler: Can do, I waffled on this one, originally writing it as you suggest, so it will be easy to switch it.

gateway/handlers/createhandler.go
LucasRoesler: must have gotten here while I was resolving a merge

api-docs/swagger.yml
LucasRoesler: Done

guide/secure_secret_management.md
LucasRoesler: I can remove them completely, keep the guide simple.  I have been directly using the api a lot, so it is where my first examples come from. The cli doesn't always expose all of the features of the gateway right away (or at least this was true in the past)

guide/secure_secret_management.md
LucasRoesler: I was able to reproduce using your curl command as well
```
➜  curl -H "Content-Type: application/json" -X POST -H "X-Api-Key: R^Y$q" -d '{}' http://localhost:8080/function/protectedapi
Unlocked the function!
```

That is really weird, the handler is fairly simple

```go
func handle(body []byte) {
	key := os.Getenv("Http_X_Api_Key")

	secretBytes, err := ioutil.ReadFile("/run/secrets/secret_api_key")
	if err != nil {
		log.Fatal(err)
	}

	secret := strings.TrimSpace(string(secretBytes))

	if key == secret {
		fmt.Println("Unlocked the function!")
	} else {
		fmt.Println("Access denied!")
	}
}
```

it seems like `golang` is doing some sort of regex match instead

LucasRoesler: Odd, if you test the string the go playground, it works as expected https://play.golang.org/p/EjWkzE1wXd

LucasRoesler: This example includes the byte to string conversion and the trim: https://play.golang.org/p/yCHDYNPlkC but it still works as expected

LucasRoesler:
@johnmccabe I agree that you wouldn't normally use secrets like this to protect an API, Kong would be better for that, but I think most of the other examples where you would use this (DB password, or 3rd party API key) require a lot more setup to show the usage and might make the guide more difficult for people to use.  Also, in this case, I was simply reusing an example that we already had.  I could add a link to the Kong guide and recommendation to use it instead for running a secure API.

guide/secure_secret_management.md
LucasRoesler: Just found the "bug", the `$` is trying to do variable substitution in bash, so the secret actually looks like `R^Y` inside the container and the key in the curl command `R^Y$q` also looks like `R^Y`

LucasRoesler:
I just pushed changes to address the comments on the previous diff

udondan:
I tested your PR, had no issues. Secrets are available in functions when deployed with `--secret foo`.  :thumbsup:

alexellis:
I need to catch up on this and understand the equivalent for Kubernetes. Thank you for the work here. Would you be able to demo in the next contributor meeting? 22/10

LucasRoesler:
I would be willing to do a demo, but I need to know the time of the meeting, I never received an actual invite.

sample-functions/ApiKeyProtected/README.md
```diff
@@ -1,6 +1,6 @@
 ### Api-Key-Protected sample
 
-To use this sample provide an env variable for the container/service in `secret_api_key`.
+See the [secure secret management guide](../guide/secure_secret_management.md) for instructions on how to use this function.
```
ericstoekl: Are there plans to create this file or does it exist somewhere else?

ericstoekl:
Tested the happy path of this one, and it works. Loaded up a node function that responds to webhooks, and it properly gets the secret that I created. Great work @LucasRoesler

sample-functions/ApiKeyProtected/README.md
alexellis: We should fix the link @LucasRoesler

gateway/handlers/createhandler.go
```diff
@@ -167,3 +179,65 @@ func userPasswordFromBasicAuth(basicAuthB64 string) (string, string, error) {
...
+	// the spec
+	args := filters.NewArgs()
+	for _, opt := range secretOpts.Value() {
+		args.Add("name", opt.SecretName)
+	}
+
+	secrets, err := c.SecretList(ctx, types.SecretListOptions{
+		Filters: args,
+	})
+	if err != nil {
```
alexellis: Should we handle this and make a 404 go back to the client?

gateway/handlers/createhandler.go
```diff
@@ -66,7 +77,7 @@ func MakeNewFunctionHandler(metricsOptions metrics.MetricOptions, c *client.Clie
 	}
 }
 
-func makeSpec(request *requests.CreateFunctionRequest, maxRestarts uint64, restartDelay time.Duration) swarm.ServiceSpec {
+func makeSpec(request *requests.CreateFunctionRequest, maxRestarts uint64, restartDelay time.Duration, secrets []*swarm.SecretReference) swarm.ServiceSpec {
```
alexellis: @LucasRoesler this must work for create and update deployment before it can be merged.

To update via CLI:

```
faas-cli update --replace=false --update=true
```

guide/secure_secret_management.md
alexellis: So how do we fix that?

gateway/handlers/createhandler.go
LucasRoesler: I am going ot push a change that fixes the `updateSpec` any opinion on if I should move the `makeSecretsArray` method to be in a central location?  It is currently in the `createhandler.go` file but will also be used in `update_handler.go`

gateway/handlers/createhandler.go
LucasRoesler: That is the responsibility of the caller, which we currently do 

```go
		secrets, err := makeSecretsArray(c, request.Secrets)
		if err != nil {
			log.Println(err)
			w.WriteHeader(http.StatusBadRequest)
			w.Write([]byte("Deployment error: " + err.Error()))
			return
		}
```

guide/secure_secret_management.md
LucasRoesler: I don't think we can "fix" it because it is the terminal doing the variable substitution,  by the time the value is accessible to any part of `faas` or `docker`, it has already been modified.   I updated the walkthrough already to remove the `$` that causes the issue.  I think the only other thing we could do is warn people about using `$` in a secret when creating the secret using `echo` in the terminal.  Basically, don't use `$` if you plan on expressing the value in the terminal _or_ ensure that you escape the `$` correctly.

LucasRoesler:
I pushed the change for the `updateHandler` and rebased it on `master` to make sure there were no conflicts or other issues.

gateway/handlers/createhandler.go
alexellis: What happens if you do this out of order:

* Create function (asking for secret1)
* Create secret1 afterwards

Does the function eventually pick up the secret?

alexellis:
Hi Lucas, John and I were reviewing the PR again. Let's circle back on this and get it re-tested for create/update and try to merge this PR next.

alexellis:
@johnmccabe

gateway/handlers/createhandler.go
LucasRoesler: I am pretty sure the secret must exist first, for example if you take the command from [the walk through on docker secrets](https://docs.docker.com/engine/swarm/secrets/#simple-example-get-started-with-secrets) 

```sh
docker service  create --name redis --secret my_secret_data redis:alpine
```
without having first create the `my_secret_data`   it replies with 

```
secret not found: my_secret_data
```

So, secrets must be created before functions, otherwise docker will throw an error.

gateway/requests/requests.go
```diff
@@ -20,6 +20,9 @@ type CreateFunctionRequest struct {
 	// EnvVars provides overrides for functions.
 	EnvVars map[string]string `json:"envVars"`
 
+	// Secrets is a list of secrets required for the orchestration layer to provide
+	Secrets []string `json:"secrets"`
```
ericstoekl: Duplicate of the `Secrets` property. It is defined below at line 35

gateway/handlers/createhandler.go
```diff
@@ -66,13 +77,19 @@ func MakeNewFunctionHandler(metricsOptions metrics.MetricOptions, c *client.Clie
...
 	constraints := []string{}
 	if request.Constraints != nil && len(request.Constraints) > 0 {
 		constraints = request.Constraints
 	} else {
 		constraints = linuxOnlyConstraints
 	}
+	labels := map[string]string{"function": "true"}
+	if request.Labels != nil {
+		for k, v := range request.Labels {
```
ericstoekl: Should be:

```
for k, v := range *request.Labels {
```

gateway/handlers/createhandler.go
```diff
@@ -66,13 +77,19 @@ func MakeNewFunctionHandler(metricsOptions metrics.MetricOptions, c *client.Clie
...
+			labels[k] = v
+		}
+	}
 
 	labels := map[string]string{
```
ericstoekl: `labels = map[string]string{`

gateway/handlers/createhandler.go
```diff
@@ -104,14 +121,19 @@ func makeSpec(request *requests.CreateFunctionRequest, maxRestarts uint64, resta
...
+				Image:   request.Image,
+				Labels:  labels,
+				Secrets: secrets,
 			},
 			Networks: nets,
 			Placement: &swarm.Placement{
 				Constraints: constraints,
 			},
 		},
+		Annotations: swarm.Annotations{
```
ericstoekl: Repeat of code already present lines 113-116

LucasRoesler: Sorry about that, those repeat lines must have snuck in we I rebased it last week.

alexellis:
No worries. Thanks to you both for working on this.

guide/secure_secret_management.md
```diff
@@ -0,0 +1,110 @@
...
+  protectedapi:
+    lang: Dockerfile
+    skip_build: true
+    image: functions/api-key-protected:latest
+```
+
+Note that unlike the `envVars` in the first example, we do not provide the secret value, just a list of names: `"secrets": ["secret_api_key"]`. The secret value has already been securely stored in the Docker swarm.  One really great result of this type of configuration is that you can simplify your function code by always referencing the same secret name, no matter the environment, the only change is how the environments are configured.
+
+## Advanced Swarm Secrets
+For various reasons, you might add a secret to the Swarm under a different name than you want to us in your function, e.g. if you are rotating a secret key. The Docker Swarm secret specification allows us some advanced configuration of secrets [by supplying a comma-separated value specifying the secret](https://docs.docker.com/engine/reference/commandline/service_create/#create-a-service-with-secrets).  The is best show in an example. Let's change the api key on our example function.
```
ericstoekl: >under a different name than you want to *use* in your function

alexellis:
Is this ready @LucasRoesler @ericstoekl ?

ericstoekl:
I tested this and it is good to go.

LucasRoesler:
I am pushing a fresh rebase and a update for the typo in the guide

LucasRoesler:
Just pushed a fresh rebase the corrects the merge conflict

guide/secure_secret_management.md
```diff
@@ -0,0 +1,110 @@
...
+
+functions:
+  protectedapi:
+    lang: Dockerfile
+    skip_build: true
+    image: functions/api-key-protected:latest
+```
+
+Note that unlike the `envVars` in the first example, we do not provide the secret value, just a list of names: `"secrets": ["secret_api_key"]`. The secret value has already been securely stored in the Docker swarm.  One really great result of this type of configuration is that you can simplify your function code by always referencing the same secret name, no matter the environment, the only change is how the environments are configured.
+
```
BurtonR: I haven't been able to get the secrets to work from the yml file... Here, you describe adding a "list of names", but there's no mention of it in the example above. 
I've tried a yml style list:
```yaml
functions:
  secret_test:
    lang: node
    handler: ./secret_test
    image: secret_test:test
    secrets: 
      - API_SECRET_KEY
```

I've also tried matching your inline example format:
```yaml
functions:
  secret_test:
    lang: node
    handler: ./secret_test
    image: secret_test:test
    secrets: ["API_SECRET_KEY"]
```
_with and without the quotes just in case_

both with no luck.

Deploying using the cli _did_ work though
`faas-cli deploy -f secret_test.yml --secret API_SECRET_KEY`

or, am I misreading this in that you don't define the secret in the yml file, but only through the cli?

LucasRoesler: Your example yml files are definitely what is intended and I should update it. When you say deploying with the CLI worked, do you mean that `faas-cli deploy -f secret_test.yml` did not work?

BurtonR: Right. 
So, setting the secrets _only_ using the yml file and deploying with the cli `faas-cli deploy -f secret_test.yml` I was not able to read the secret. This is the same with deploying via the UI (no secret)

Deploying with the cli adding the additional secrets flag `faas-cli deploy -f secret_test.yml --secret API_SECRET_KEY` did work and the secret is available

LucasRoesler: Seems like it might be an error in the `fass-cli` parsing the stack file, I will take another look

ericstoekl: @BurtonR , are you updated to the latest version of the faas-cli? The secrets parsing was added fairly recently

LucasRoesler: Looks like release `0.5.1` should support both the stack file and the cli args

ericstoekl: For the record, I just tried this out, deployed the `apikeyprotected` function with the secret specified in the stack yaml, and the secret is showing up in the container (at `/run/secrets/<secretname>`). @BurtonR , are you seeing an issue with the secret not showing up in the container at all, or are you seeing the secret show up in the container but not be able to read the secret from its location in `/run/secrets/`?

BurtonR: @ericstoekl not sure what the difference is between being in the container, and being in the secrets location...

When testing this, I created a simple node function to read the secret from the `/run/secrets/` location and it was throwing an error that the file didn't exist (the secret wasn't available in the function)

Also, the cli 0.5.1 hasn't been released yet. It's still marked as Pre-Release, which is why I don't have it yet. I'll grab the binary and try this out again. 

The cli version compatibility should be a note in the guide 😉

BurtonR: Alright, so, using the faas-cli version 0.5.1 with the secrets defined in the yml file:
```yaml
functions:
  secret_test:
    lang: node
    handler: ./secret_test
    image: secret_test:test
    secrets: 
      - API_SECRET_KEY
```

The secret _is_ available and working as I expect it to.

Once the guide is updated to include these couple of little things I'd call this good to go! 
Nice work @LucasRoesler!

LucasRoesler: I have updated the guide with a note at the beginning of the section using Swarm secrets.

alexellis:
Thanks Lucas for working so hard on this and to others for collaborating.

guide/secure_secret_management.md
alexellis: OK 👍

guide/secure_secret_management.md
alexellis: Thanks for testing this out @BurtonR

sample-functions/NodeHelloEnv/handler.js
```diff
@@ -0,0 +1,13 @@
+"use strict"
```
alexellis: What is this sample for?

gateway/vendor.conf
alexellis: What change needed the new vendored component?

alexellis:
Tested this briefly on my lunch break today. It appears to work well 🎉 

* Added secret to url-ping - found it via `docker exec`
* Did a rolling update `--update=true --replace--false` and found the secret got removed as expected.

gateway/vendor.conf
LucasRoesler: This was for processing secrets from strings, this allows for using the same style of arguments that the cli accepts.  For example 

```
source=mysql_password,target=wp_db_password,mode=0400
```
like they document at https://docs.docker.com/engine/swarm/secrets/#advanced-example-use-secrets-with-a-wordpress-service

sample-functions/NodeHelloEnv/handler.js
LucasRoesler: I built this one to have a separate sample function to demonstrate accessing the ENV versus Secrets

sample-functions/ApiKeyProtected/Dockerfile.build
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.7.5
```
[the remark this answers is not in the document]
LucasRoesler: Just pushed an update that includes fixes for both sample functions

sample-functions/NodeHelloEnv/handler.js
alexellis: Just trying to work out if we still have the origin "ApiKeyProtected" example which used env-vars?

LucasRoesler: I originally created this example for ENV and then update `ApiKeyProtected` to use secrets. I was asked to put the original `ApiKeyProtected` back and create a separate sample for secrets.

alexellis:
Merged. Thank you @LucasRoesler - we need to get the community to test the new release 0.16.15.

PR #299, opened 2017-10-15
Add start, stop, and status commands to deploy scripts

markpeek:
Signed-off-by: Mark Peek <markpeek@vmware.com>

## Description
This change adds start, stop, and status commands to the deploy_*.sh scripts.

## Motivation and Context
Improve the ease of use for new users with an easy way to both start and stop the stack.

## How Has This Been Tested?
Manual testing of the scripts except for the armhf script.

### DIFF
--- a/deploy_stack.armhf.sh
+++ b/deploy_stack.armhf.sh
@@ -5,5 +5,32 @@ if ! [ -x "$(command -v docker)" ]; then
   exit 1
 fi
 
-echo "Deploying stack"
-docker stack deploy func --compose-file docker-compose.armhf.yml
+if [ $# -eq 0 ]; then
+  command="start"
+else
+  command="$1"
+fi
+
+case "$command" in
+  start)
+    echo "Deploying stack"
+    docker stack deploy func --compose-file docker-compose.armhf.yml
+    ;;
+  stop)
+    echo "Stopping stack"
+    docker stack rm func
+    ;;
+  status)
+    docker stack ps func
+    ;;
+  help)
+    echo "Available commands:"
+    echo "start  - start the stack"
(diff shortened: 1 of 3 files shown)


deploy_stack.armhf.sh
```diff
@@ -5,5 +5,32 @@ if ! [ -x "$(command -v docker)" ]; then
...
+fi
+
+case "$command" in
+  start)
+    echo "Deploying stack"
+    docker stack deploy func --compose-file docker-compose.armhf.yml
+    ;;
+  stop)
+    echo "Stopping stack"
+    docker stack rm func
```
johnmccabe: The start/stop/status commands make this look like a service - but the stop here is gonna nuke the stack rather than pause it. Is that what was intended?
Gut feel here is that its confusing the purpose of the script - which from its name is to deploy.
Perhaps generic `generic stop_stack.sh` and `status_stack.sh` scripts instead - they don't need variants I suspect.

markpeek: Agreed, in some ways the deploy script is bringing up a service and the other commands are helping to manage it. Yes, the stop was intended to nuke it...other command name suggestions? And I preferred to put the functionality into a single script given these scripts are at the top level of the repo (repo pollution).

alexellis:
Hi @markpeek,

Thank you for taking the time to work on OpenFaaS. To John's comment - I think might need to spend a bit more time thinking about how to do this.

alexellis:
@markpeek once you have a moment or are back home, please can we open an issue to talk about the design?

alexellis:
Ping @markpeek

alexellis:
Closing due to inactivity. Feel free to propose a new change via an issue.

Issue #234, opened 2017-09-28
Proposal: Kafka queue-worker


alexellis:
In a similar vein to [openfaas/nats-queue-worker](https://github.com/openfaas/nats-queue-worker), we should support Kafka (initially a set topic/partition) as a back-end for asynchronous process and model the NATS Streaming implementation.

Notes:
* Kafka runs its own cluster with zookeeper as an additional dependency
* Kafka complexity is several times greater than NATS Streaming
* Needs a new repo i.e. openfaas/kafka-queue-worker
* Changes to main.go/server.go in the faas repo
* Changes to the async-function route handler

alexellis:
Derek add label: hacktoberfest

johnmccabe:
@kinghuang is this something you'd be interested in

VitorFalcao:
@alexellis Could I work on this? If yes, could you give me some pointers? This seems to be an awesome feature.

alexellis:
Stay tuned.. we may have something from one our contributors coming around this

StevenACoffman:
So, kubeless integrates kafka. It's also written in golang. Is there any interest in collaborating?

yglcode:
as an learning exercise, i added initial code to use kafka, along the path of minimal impact:

* follow alex's suggestions, current code/flow for nats queue worker, add code in gateway and add kafka_queue_worker repository.

* add docker-compose.kafka-queue.yml to bring up zookeeper & kafka nodes in swarm together with gateway and kafka-queue-worker.

* to avoid tool-chain change, use Shopify sarama package instead of confluent-kafka-go package. while the latter quarantees the latest features, it introduces dependencies on C/C++ librdkafka library. we may not be able to build static linked binaries and have to use [alpine images with librdkafka built in](http://github.com/yglcode/alpine-kafka-go).

links:
[kafka_queue_worker branch](http://github.com/yglcode/faas).
[kafka-queue-worker repo](http://github.com/yglcode/kafka-queue-worker).

any suggestions? i am happy to help share and merge the code.

alexellis:
Thank you for all the work on this. It was used in the Kafka-connector, but the gateway still uses NATS Streaming for Async. We may revisit in the future depending on appetite from users.

PR #311, opened 2017-10-17
Kafka queue worker

yglcode:
kafka-queue-worker: this PR cover the changes at Gateway.

## Description
initial code to use kafka, along the path of minimal impact:

* follow alex's suggestions, current code/flow for nats queue worker, add code in gateway and add kafka_queue_worker repository.

* add docker-compose.kafka-queue.yml to bring up zookeeper & kafka nodes in swarm together with gateway and kafka-queue-worker.

* to avoid tool-chain change, use Shopify sarama package instead of confluent-kafka-go package. while the latter quarantees the latest features, it introduces dependencies on C/C++ librdkafka library. we may not be able to build static linked binaries

## Motivation and Context

Support use cases to use kafka cluster as backing async work queue, similar to NATS in a different repo.

[related open issue](https://github.com/openfaas/faas/issues/234)

Alex originally raised this issue to propose the change.

## How Has This Been Tested?

Test with local swarm cluster with 2-3 kafka brokers + 1 zookeeper.

Tests runs in local swarm cluster. tested by running simple scripts similar to:
```
for i in {1..10};do curl http://$SWAMGRIP:8080/async-function/kk_echoit --data "hi";done
```
Follow closely current code base for NATS, minimal impact to other parts.

## Types of changes

## Checklist:


alexellis:
Thank you for this. It would be good for me to get some input from @rawkode and @kinghuang too who also work with Kafka. 👍

alexellis:
See also - https://github.com/yglcode/kafka-queue-worker

yglcode:
These dependencies are what "vndr" automatically pulls in when running against kafka-queue-worker/queue-worker; Basically it includes Shopify/sarama and bsm/sarama-cluster, and their dependencies. You are right that at gateway we only use sarama's producer api, so no need for sarama-cluster. I can rerun "vndr" without sarama-cluster which may remove dependencies for gateway.

alexellis:
Let's try the re-vendoring and see what we get 👍 Thank you

docker-compose.kafka-queue.yml
```diff
@@ -0,0 +1,189 @@
+version: "3.2"
+services:
+    gateway:
+        volumes:
+            - "/var/run/docker.sock:/var/run/docker.sock"
+        ports:
+            - 8080:8080
+        image: ${REGISTRY_SLASH}functions/gateway${COLON_ETAG}
```
alexellis: I think you can specify a default here too in the expression

docker-compose.kafka-queue.yml
```diff
@@ -0,0 +1,189 @@
+version: "3.2"
+services:
```
alexellis: I've looked at this config for the first time, it's huge. Is there a way of putting it in another file? We'd also need this in the Kubernetes YAML format in the faas-netes repo.

docker-compose.kafka-queue.yml
```diff
@@ -0,0 +1,189 @@
...
+                max_attempts: 20
+                window: 380s
+            placement:
+                constraints:
+                    - 'node.platform.os == linux'
+
+    # End
+
+    prometheus:
+        image: ${REGISTRY_SLASH}functions/prometheus:latest  # autobuild from Dockerfile in repo.
```
alexellis: Do we need this prefix on every container? What was it added for?

gateway/server.go
```diff
@@ -118,6 +119,16 @@ func main() {
 
 		faasHandlers.QueuedProxy = internalHandlers.MakeQueuedProxy(metricsOptions, true, &logger, natsQueue)
 		faasHandlers.AsyncReport = internalHandlers.MakeAsyncReport(metricsOptions)
+	} else if config.UseKafka() {
```
alexellis: Do you think a Golang 1.9 plugin can help here now that we have two providers?

alexellis:
> yglcode Let's try the re-vendoring and see what we get 👍 Thank you

@yglcode wanted to ping you about this PR. I've left comments

yglcode:
after re-vendoring, it seems that only bsm/sarama-cluster is removed.
${REGISTRY_SLASH} and ${COLON_TAG} added to customize for local dev/test to use images in local registry. In other environments, without definition, these should resolve to empty space, so will fetch from dockerhub by default. However as you mentioned, add a default value for ${COLON_TAG} could be better.
Yes, a separate config file for zookeer& kafka can be used, if that fits better.

grillz:
Forgive my ignorance, but with this would it be possible to have functions that subscribe to certain topics? i.e. I have a `user.updated` topic that I want to functions to consume from.

alexellis:
Yes but this is a different piece of work. We have someone on Slack asking about a Kafka connector and I pointed him at this... https://github.com/ucalgary/ftrigger from the University of Calgary. If you'd like to join Slack and find out more ping me alex@openfaas.com

alexellis:
It'd also be good to remove the syntax for `${COLON_TAG}`  since we don't use it anywhere else yet.

alexellis:
Let's separate out all the testing Kakfa setup files into guide / separate repo. I think people will generally have Kafka up and running the way they like it. It also makes the PR easier to merge + review.

yglcode:
kafka test setup are moved to a separate file docker-compose.kafka-setup.yml.
Since in docker-compose.kafka-queue.yml, the settings for "kafka-queue-worker" and "addtopics" refer to the names of "zk1","kafka1,2...", is it better to keep these two files together?
Regarding to go1.9 plugin, the concern is that its implementation not complete, only available for linux. right now "gateway" is a nice fully linked binary which can run cross-arch cross-platform which is a big plus.

alexellis:
Thank you for working on our feedback. I think we can avoid the plugin for now then. I would suggest that we move the config files for Kakfa into the contrib folder.

docker-compose.kafka-queue.yml
```diff
@@ -0,0 +1,110 @@
...
+        ports:
+            - 8080:8080
+        image: functions/gateway:latest-dev
+        networks:
+            - functions
+        environment:
+            dnsrr: "true"  # Temporarily use dnsrr in place of VIP while issue persists on PWD
+            faas_kafka_brokers: kafka1:19092,kafka2:19092
+            faas_queue_topics: faas-request # ,other-topics
+        depends_on:
```
alexellis: Please remove the "depends_on" setting since we just found out it's deprecated anyway.

docker-compose.kafka-queue.yml
```diff
@@ -0,0 +1,110 @@
...
+                delay: 5s
+                max_attempts: 20
+                window: 380s
+            placement:
+                constraints:
+                    - 'node.platform.os == linux'
+
+    # End
+
+    prometheus:
```
alexellis: I wonder if we run one file after another, whether that can avoid having to repeat all the other services that are the same in both files?

docker-compose.kafka-setup.yml
```diff
@@ -0,0 +1,83 @@
+version: "3.2"
+services:
+    # zookeeper
+    zk1:
+        image: confluentinc/cp-zookeeper:3.3.0
```
alexellis: For a development environment there is apparently a Go-binary which is compatible with Kafka, it may be better? Most people who opt to use Kakfa will deploy it themselves or have it set up already.

alexellis:
I'd also like to get input from @kinghuang who is using Kakfa for event triggers. @yglcode I also wanted to point you at a new issue we have in #361

alexellis:
Thank you for working on this, it's an important piece of the project. Comments added 4 days ago. Are you in our Slack group yet?

yglcode:
hi alex, sorry to reply late. I started busy with something else (possible from now on).
move kafka setup file to contrib/ and remove "depends_on".
regarding to reuse service config, it seems that we cannot just start kafka after the default config, since we need customize "gateway" with env vars for kafka brokers, similar to nats. and docker has pull/patch adding support for "docker stack deploy" with multiple docker-compose files, so we can wait later to do: docker stack deploy -c docker-compose.yml -c docker-compose.kafka-worker.yml.

yglcode:
also what is the Go-binary you mentioned compatible with kafka? i am aware of "jocko", but jocko just adding replication and miss consumer-group, i am not sure if it can help us or not.

alexellis:
Please could you update to use all standard ports?

yglcode:
please feel free to take over the code and fixed it as needed.

alexellis:
I contacted you on the Slack channel, don't know if you've logged in for a while?

alexellis:
Thank you for all your time on this, it's very good work, just want to make sure we can use it in a way that is swappable and componentised. 

I took your research as a basis to start writing a Go connector for the API gateway and Kafka. https://github.com/openfaas-incubator/kafka-connector

Would be good to get your feedback on that too.

alexellis:
@yglcode pinging :+1:

yglcode:
the code looks good to me.:)

in my limited understanding, it looks quite different from ftrigger from ucalgary. while ftrigger transparently bind kafka topic to faas functions thru swarm service labels and without any work from kafka/app side; this connector will require users setting up "faas-request" topic in their kafka cluster, and app code sending requests {function-name, data} to this topic.

it is close to how current "async-function" works, if i got it right; the difference is that in "connector", work-queue is outside of faas, maintained by user, sitting in front of gateway, while in "async-function" api, worker-queue is managed by faas (hidden from user), sitting behind gateway.

alexellis:
The concern was dynamically adding and removing consumers for topics. I wasn't sure how it could be done well with the library - so then looked to build around a single topic to see if it would still serve the purpose.

The queue worker is hidden that's correct through the asynchronous route.

yglcode:
i guess both ftrigger and "connector" have pros & cons. could use both.

alexellis:
@yglcode Thank you for your work. This PR has really helped me understand how to build a connector/integration that also avoids using CGO. I'm not planning to merge this approach, but to look to offer Kafka as an external integration like the kafka-connector I shared.

I'll close the PR and mark it #revisit, if you'd like to help with the kafka-connector please let me know.

Issue #273, opened 2017-10-09
Research: Report function replica status by querying Docker Swarm API


ericstoekl:
From Alex:
> Do we have a way in the Docker Swarm API to query whether any of the 1/1 replicas are “ready” i.e. healthcheck passed? This could help us in making our error messages better.. or if there’s another state like “pending” that we can return as “node pulling image” etc

We need a way to report to the user whether their function is ready-to-go, without making them examine the output of `docker service ls` (and probably `| grep <functionname>`), which is currently what's required to find out whether your container is replicated or not.

Each of the different backends (Swarm, K8, etc) will need their own implementation of this feature, because each backend manages the health of their containers in a different way. We will first get this working with Docker Swarm.

## Expected Behaviour

When using `proxy.ListFunctions()` in the `faas-cli` code, it will return info about what functions are available, and whether or not they are ready to use.

## Current Behaviour

The CLI tool is unable to report whether or not a function is ready to use.

## Possible Solution

We will add this feature to the `gateway/handlers/` section. Need to read through the Docker Swarm API to figure out exactly how to do this.

## Steps to Reproduce (for bugs)

1. Deploy a new function with `faas-cli deploy -f <YAML>`
2. In `faas-cli list`, you will see the function name appear, even before it is active. However, there is no indication of whether the function is ready to use or not.
3. Need to query `docker service ls` to see if the function has any active replicas or not (wait until replicas status is `1/1`)
4.

## Context

This is an issue that we have been talking about for some time. The contention is that separate functionality will be needed for each backend. However, we would like to start with Docker Swarm.

## Your Environment

* Docker version `docker version` (e.g. Docker 17.0.05 ):

* Are you using Docker Swarm or Kubernetes (FaaS-netes)?
Docker Swarm

* Operating System and version (e.g. Linux, Windows, MacOS):
Targeting all OSes

* Link to your project or a code example to reproduce issue:

ericstoekl:
Derek assign issue to: ericstoekl

ericstoekl:
Derek assign: ericstoekl

johnmccabe:
Docker add label: proposal

johnmccabe:
Derek add label: proposal

ericstoekl:
Update: https://asciinema.org/a/M7OS28J4tvza0oe7ZVXLusfqa

Will be making a pull request soon, if this looks good.

ericstoekl:
[...]

PR #315, opened 2017-10-20
[WIP] Report Function readiness state (Swarm)

ghost:
```
Report function replica status

1) Add getReplicaInfo() function to handlers/reader.go, which allows for
reporting of number of active functions
2) Add verbose mode for getting this info.
```

## Description

Look for a parameter called `v` in the `r.URL` data from the request. If it exists, set the `verbose` param to true, which will enable calling of `getReplicaInfo()`, which will go through the process of getting the health of all instances.

`getReplicaInfo()` code is similar to the code from the docker CLI `runList()` function: https://github.com/docker/cli/blob/master/cli/command/service/list.go

## Motivation and Context

We have a research item where we have been tracking this issue: #273 

## How Has This Been Tested?

Added tests that check response code for success, and whether or not available replica count was acquired, with and without `--verbose` flag.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/handlers/reader.go
+++ b/gateway/handlers/reader.go
@@ -59,19 +59,19 @@ func MakeFunctionReader(metricsOptions metrics.MetricOptions, serviceClient clie
 				// Required (copy by value)
 				labels := service.Spec.Annotations.Labels
 
-				var replicaCount int
+				var availableReplicas int
 				if _, ok := running[service.ID]; ok {
-					replicaCount = running[service.ID]
+					availableReplicas = running[service.ID]
 				}
 
 				f := requests.Function{
-					Name:            service.Spec.Name,
-					Image:           service.Spec.TaskTemplate.ContainerSpec.Image,
-					InvocationCount: 0,
-					Replicas:        *service.Spec.Mode.Replicated.Replicas,
-					ReplicaCount:    replicaCount,
-					EnvProcess:      envProcess,
-					Labels:          &labels,
+					Name:              service.Spec.Name,
+					Image:             service.Spec.TaskTemplate.ContainerSpec.Image,
+					InvocationCount:   0,
+					Replicas:          *service.Spec.Mode.Replicated.Replicas,
+					AvailableReplicas: availableReplicas,
+					EnvProcess:        envProcess,
+					Labels:            &labels,
 				}
 
 				functions = append(functions, f)
--- a/gateway/handlers/reader.go
+++ b/gateway/handlers/reader.go
@@ -13,22 +13,29 @@ import (
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/swarm"
 	"github.com/docker/docker/client"
 	"github.com/openfaas/faas/gateway/metrics"
 	"github.com/openfaas/faas/gateway/requests"
 )
 
 // MakeFunctionReader gives a summary of Function structs with Docker service stats overlaid with Prometheus counters.
-func MakeFunctionReader(metricsOptions metrics.MetricOptions, c client.ServiceAPIClient) http.HandlerFunc {
+func MakeFunctionReader(metricsOptions metrics.MetricOptions, serviceClient client.ServiceAPIClient, nodeClient client.NodeAPIClient) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 
+		verbose := false
+
+		if r.URL != nil {
+			verbose = queryIsNotFalse(r, "v")
+		}
 		serviceFilter := filters.NewArgs()
 
 		options := types.ServiceListOptions{
 			Filters: serviceFilter,
 		}
 
-		services, err := c.ServiceList(context.Background(), options)
+		ctx := context.Background()
+		services, err := serviceClient.ServiceList(ctx, options)
 		if err != nil {
 			log.Println("Error getting service list:", err)
 			w.WriteHeader(http.StatusInternalServerError)
@@ -39,6 +46,11 @@ func MakeFunctionReader(metricsOptions metrics.MetricOptions, c client.ServiceAP
 		// TODO: Filter only "faas" functions (via metadata?)
 		functions := []requests.Function{}
 
+		var running map[string]int
+		if verbose {
+			running = getReplicaInfo(serviceClient, nodeClient, services, ctx)
+		}
+
 		for _, service := range services {
 
 			if len(service.Spec.TaskTemplate.ContainerSpec.Labels["function"]) > 0 {
@@ -47,11 +59,17 @@ func MakeFunctionReader(metricsOptions metrics.MetricOptions, c client.ServiceAP
 				// Required (copy by value)
 				labels := service.Spec.Annotations.Labels
 
+				var replicaCount int
+				if _, ok := running[service.ID]; ok {
+					replicaCount = running[service.ID]
+				}
+
 				f := requests.Function{
 					Name:            service.Spec.Name,
 					Image:           service.Spec.TaskTemplate.ContainerSpec.Image,
 					InvocationCount: 0,
 					Replicas:        *service.Spec.Mode.Replicated.Replicas,
+					ReplicaCount:    replicaCount,
 					EnvProcess:      envProcess,
 					Labels:          &labels,
 				}
@@ -77,4 +95,42 @@ func getEnvProcess(envVars []string) string {
 	}
 
 	return value
+}
+
+func getReplicaInfo(serviceClient client.ServiceAPIClient, nodeClient client.NodeAPIClient, services []swarm.Service, ctx context.Context) map[string]int {
+	// Begin replica info section
+	taskFilter := filters.NewArgs()
+
+	taskFilter.Add("_up-to-date", "true")
+	tasks, err := serviceClient.TaskList(ctx, types.TaskListOptions{Filters: taskFilter})
+	if err != nil {
+		log.Println(err)
+	}
+
+	nodes, err := nodeClient.NodeList(ctx, types.NodeListOptions{})
+	if err != nil {
+		log.Println(err)
+	}
+
+	activeNodes := make(map[string]struct{})
+	for _, n := range nodes {
+		if n.Status.State != swarm.NodeStateDown {
+			activeNodes[n.ID] = struct{}{}
+		}
+	}
+
+	running := map[string]int{}
+	for _, task := range tasks {
+		if _, nodeActive := activeNodes[task.NodeID]; nodeActive && task.Status.State == swarm.TaskStateRunning {
+			running[task.ServiceID]++
+		}
+	}
+
+	//log.Printf("running: %v\n", running)
+	return running
+}
+
+func queryIsNotFalse(r *http.Request, k string) bool {
+	s := strings.ToLower(strings.TrimSpace(r.FormValue(k)))
+	return !(s == "" || s == "0" || s == "no" || s == "false" || s == "none")
 }
--- a/gateway/requests/requests.go
+++ b/gateway/requests/requests.go
@@ -51,12 +51,12 @@ type FunctionResources struct {
 
 // Function exported for system/functions endpoint
 type Function struct {
-	Name            string  `json:"name"`
-	Image           string  `json:"image"`
-	InvocationCount float64 `json:"invocationCount"` // TODO: shouldn't this be int64?
-	Replicas        uint64  `json:"replicas"`
-	ReplicaCount    int     `json:"replicaCount"`
-	EnvProcess      string  `json:"envProcess"`
+	Name              string  `json:"name"`
+	Image             string  `json:"image"`
+	InvocationCount   float64 `json:"invocationCount"` // TODO: shouldn't this be int64?
+	Replicas          uint64  `json:"replicas"`
+	AvailableReplicas int     `json:"availableReplicas"`
+	EnvProcess        string  `json:"envProcess"`
 
 	// Labels are metadata for functions which may be used by the
 	// back-end for making scheduling or routing decisions
--- a/gateway/requests/requests.go
+++ b/gateway/requests/requests.go
@@ -55,6 +55,7 @@ type Function struct {
 	Image           string  `json:"image"`
 	InvocationCount float64 `json:"invocationCount"` // TODO: shouldn't this be int64?
 	Replicas        uint64  `json:"replicas"`
+	ReplicaCount    int     `json:"replicaCount"`
 	EnvProcess      string  `json:"envProcess"`
 
 	// Labels are metadata for functions which may be used by the
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -83,7 +83,7 @@ func main() {
 
 		faasHandlers.Proxy = internalHandlers.MakeProxy(metricsOptions, true, dockerClient, &logger)
 		faasHandlers.RoutelessProxy = internalHandlers.MakeProxy(metricsOptions, false, dockerClient, &logger)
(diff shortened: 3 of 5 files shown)


alexellis:
Derek add label: design/review

docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
             - "/var/run/docker.sock:/var/run/docker.sock"
         ports:
             - 8080:8080
-        image: functions/gateway:0.6.6-beta
+        image: functions/gateway:latest-dev
```
alexellis: This might be incorrect btw.

alexellis:
Hi Eric, thank you for working on this. I wanted to write a review of the replica status PR.

* What is the active nodes doing? in reader.go

* Can we filter for state active via the Swarm API rather than fetching every task that ever existed for the Service - over time this will grow slower

* verboseMode := r.URL.Query().Get("v")

I think `verbose` could be better. Would like other opinions

* MakeFunctionReader - I imagined we would pick up the services first then the tasks after.

* I also wonder if we should extend the API to query the data for a specific service ID - this seems like a common use-case and it would allow us to optimize.

* AvailableCount - Is there more value in returning an array with details of each task and its status over a simple integer?

On this note - the integer should have a name that's symmetric with "Replicas" - i.e. ReplicasReady

Additional tests seem reasonable - can we abstract this in any way so that the logic to determine status can also be tested? Look at the ServiceQuery interface I wrote for the alert scaling and how the tests operate for this.

docker-compose.yml
ericstoekl: For sure, for testing this is needed though (otherwise it will use `0.6.6-beta` which is on the remote registry)

ericstoekl:
@alexellis Thanks for your feedback, I have responded with some questions of my own:

1. What is the active nodes doing? in reader.go

Makes sure that the node that hosts the given [Task](https://godoc.org/github.com/moby/moby/api/types/swarm#Task) is on a [Node](https://godoc.org/github.com/moby/moby/api/types/swarm#Node) that is running. I suppose we could assume that the node is running, which might mean we can remove this. I'll look into it.

2. Can we filter for state active via the Swarm API rather than fetching every task that ever existed for the Service - over time this will grow slower

I am not sure if we can filter on CurrentState for the tasks at least. I've looked into it, and `swarmkit` uses a `ListTasksRequest_Filters` struct, which you can see [here](https://github.com/docker/swarmkit/blob/da5ee2a69176ca3c1eed1b04c4308bc1ab932c05/api/control.pb.go#L189). It looks like it has `DesiredStates` but not `CurrentStates`. `ListServicesRequest_Filter` has one for Runtime, but I don't think that's what we need either.

Do you know of another way to filter for CurrentState as Running?

3. I think verbose could be better. Would like other opinions

It's a HTTP GET, so this is the only way I could think of to add a parmeter to the GET request. Other option I can think of is is changing the ListHandler to a POST type.

4. MakeFunctionReader - I imagined we would pick up the services first then the tasks after.

Looks to me like we are already doing this? Or could you provide more info on this question

5. I also wonder if we should extend the API to query the data for a specific service ID - this seems like a common use-case and it would allow us to optimize.

This is something we can easily filter for

6. AvailableCount - Is there more value in returning an array with details of each task and its status over a simple integer?

Will look at this after we have more higher up items settled.

ReplicasReady sounds reasonable. Will look into more testing as well.

ericstoekl:
@alexellis - I've attempted to address the Service filtering issue we discussed earlier today.

I ended up finding the place where the filter keys are defined, it's here: https://github.com/moby/moby/blob/ab0eb8fcf6fe6b4ee12393edcd6465501b5c66a1/daemon/cluster/filters.go#L53

I saw that `"desired-state"` is another available option, so I now have the code filter for `'desired-state == running'` instead of looping through each service and adding a filter for that specific service.

Do you think this is sufficient to solve the task-count-grow-larger-over-time issue? Do you think we still need to add a `current-state` filter to swarmkit?

Thanks,
Eric

gateway/handlers/reader.go
```diff
@@ -65,3 +78,36 @@ func MakeFunctionReader(metricsOptions metrics.MetricOptions, c *client.Client)
 		w.Write(functionBytes)
 	}
 }
+
+func getReplicaInfo(c *client.Client, services []swarm.Service, ctx context.Context) map[string]int {
+	// Begin replica info section
+	taskFilter := filters.NewArgs()
+
+	taskFilter.Add("desired-state", "running")
```
alexellis: Not sure if you are filtering the services here?

gateway/handlers/reader.go
```diff
@@ -65,3 +78,36 @@ func MakeFunctionReader(metricsOptions metrics.MetricOptions, c *client.Client)
...
+	tasks, err := c.TaskList(ctx, types.TaskListOptions{Filters: taskFilter})
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	nodes, err := c.NodeList(ctx, types.NodeListOptions{})
```
alexellis: Is this necessary?

gateway/handlers/reader.go
```diff
@@ -11,31 +11,43 @@ import (
...
 
+		verbose := false
+		verboseMode := r.URL.Query().Get("v")
+		if len(verboseMode) != 0 {
+			verbose = true
+		}
 		serviceFilter := filters.NewArgs()
 
 		options := types.ServiceListOptions{
 			Filters: serviceFilter,
```
alexellis: This is not relating to your change, but we may have to optimize this. Can you add a filter here for labels or not? i.e. `function=true`

gateway/handlers/reader.go
```diff
@@ -52,6 +64,7 @@ func MakeFunctionReader(metricsOptions metrics.MetricOptions, c *client.Client)
 					Image:           service.Spec.TaskTemplate.ContainerSpec.Image,
 					InvocationCount: 0,
 					Replicas:        *service.Spec.Mode.Replicated.Replicas,
+					AvailableCount:  running[service.ID],
```
alexellis: I'm not sure this is safe, shouldn't we check if service.ID is a member of the collection first?

gateway/handlers/reader.go
```diff
@@ -11,31 +11,43 @@ import (
...
+		services, err := c.ServiceList(ctx, options)
 		if err != nil {
 			fmt.Println(err)
 		}
 
 		// TODO: Filter only "faas" functions (via metadata?)
 		var functions []requests.Function
 
+		var running map[string]int
+		if verbose {
```
alexellis: At this point I wonder if we need to add GET /system/function/<function_name> that does the verbose query if we want to look up the status of only one function from the UI or similar.

gateway/requests/requests.go
```diff
@@ -62,6 +62,7 @@ type Function struct {
 	Image           string  `json:"image"`
 	InvocationCount float64 `json:"invocationCount"` // TODO: shouldn't this be int64?
 	Replicas        uint64  `json:"replicas"`
+	AvailableCount  int     `json:"availableCount"`
```
alexellis: I might have suggested `AvailableCount` - I wonder if we need to change the name to be cohesive with the `Replicas` value such as ReplicasAvailable / ReplicasReady.

gateway/handlers/reader.go
developius: This would be useful, you'll see in my PR to @austinfrey's Node library that I had to filter the returned response from `/system/functions` to find the relevant function. I think the more RESTful the API, the better 👍

https://github.com/developius/node-openfaas/blob/e569e50b61b7190b6ec8806bb96eda1343e28437/openfaas/index.js#L23-L41

gateway/handlers/reader.go
ericstoekl: It will be `nil` if `service.ID` is not in the `running`. `AvailableCount` would then be set to `nil`. The only time `AvailableCount` is used is to print the value, so `nil` should be safe. However, it would improve readability to have an `if` statement above that checks if this value is nil, and if so set `AvailableCount` to 0.

gateway/handlers/reader.go
ericstoekl: Right you are, this is unnecessary.

gateway/handlers/reader.go
ericstoekl: I need to filter the tasks list instead of the services list. In docker swarm API, the Services list takes a different set of filters. I don't think they have a filter that will help us get running state:

https://github.com/docker/swarmkit/blob/da5ee2a69176ca3c1eed1b04c4308bc1ab932c05/api/control.pb.go#L298

gateway/handlers/reader.go
ericstoekl: Correction, I believe this may be necessary. I removed this code, then thought that I re-deployed the new image, but it turns out it was the old image. Now that I have this removed, I see that it shows every function as having replica status `1/1`, even though `docker service ls` may show the same function as `0/1`. I think filtering through the `NodeList` may be necessary

ericstoekl: Re-correction, I believe it's not necessary to loop through `nodes`.

gateway/handlers/reader.go
ericstoekl: Looks like we can filter for `labels` in both services and tasks.

Services: https://github.com/docker/swarmkit/blob/da5ee2a69176ca3c1eed1b04c4308bc1ab932c05/api/control.pb.go#L301

Tasks: https://github.com/moby/moby/blob/ab0eb8fcf6fe6b4ee12393edcd6465501b5c66a1/daemon/cluster/filters.go#L52

ericstoekl:
Updated the PR to reflect most of @alexellis's changes. I did not address [this comment](https://github.com/openfaas/faas/pull/315#discussion_r148969484) because I think it may be suited for another PR.

Consult the following Ascii Cinema for the functionality so far:

https://asciinema.org/a/P2ucBJxLAsuNmChrr7AtmStgY

The `time` command in linux tells me that with verbose, it takes about 0.01 seconds longer with `--verbose`. This may be due in part to the additional printing it is doing. I will need to test on the raspberry pi to see if cutting the unnecessary node listing has shortened the execution time.

gateway/handlers/reader.go
alexellis: That could help us with performance then - as long as the label is propagated to the tasks.

alexellis:
That looks amazing. What an improvement!

Looking forward to seeing what it's like on an RPi.

alexellis:
I also wanted to call someone from Docker to take a quick look. @thaJeztah has agreed to checkout the usage of the Swarm/Docker API. (thank you)

gateway/handlers/reader.go
```diff
@@ -71,3 +93,24 @@ func MakeFunctionReader(metricsOptions metrics.MetricOptions, c client.ServiceAP
...
+
+	taskFilter.Add("_up-to-date", "true")
+	tasks, err := c.TaskList(ctx, types.TaskListOptions{Filters: taskFilter})
+	if err != nil {
+		log.Println(err)
+	}
+
+	running := map[string]int{}
+	for _, task := range tasks {
+		if task.Status.State == swarm.TaskStateRunning {
```
thaJeztah: Wondering if node state should be taken into account; if a node is not reachable, task states are not updated (i.e., the task state then reflects the last _known_ state); see https://github.com/docker/cli/blob/master/cli/command/service/list.go#L118-L120

```go
if _, nodeActive := activeNodes[task.NodeID]; nodeActive && task.Status.State == swarm.TaskStateRunning {
	running[task.ServiceID]++
}
```

Be aware that in all cases, this state would likely be on a "best effort" base; managers (and raft store) have the _desired state_ in store, and the nodes report back what the actual state is, so there may be a latency.

gateway/handlers/reader.go
```diff
@@ -22,13 +23,22 @@ import (
 func MakeFunctionReader(metricsOptions metrics.MetricOptions, c client.ServiceAPIClient) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 
+		verbose := false
+
+		if r.URL != nil {
+			verboseMode := r.URL.Query().Get("v")
+			if len(verboseMode) != 0 {
```
thaJeztah: May want to check actual accepted values here (i.e. for the `?v=0` or `?v=false` case.

The Docker API uses a utility for this - definitely not perfect as well; I think I would've preferred a more "strict" approach, but it does the job for most values that people would "expect" to work; https://github.com/moby/moby/blob/master/api/server/httputils/form.go#L9-L22

```
// BoolValue transforms a form value in different formats into a boolean type.
func BoolValue(r *http.Request, k string) bool {
	s := strings.ToLower(strings.TrimSpace(r.FormValue(k)))
	return !(s == "" || s == "0" || s == "no" || s == "false" || s == "none")
}

// BoolValueOrDefault returns the default bool passed if the query param is
// missing, otherwise it's just a proxy to boolValue above.
func BoolValueOrDefault(r *http.Request, k string, d bool) bool {
	if _, ok := r.Form[k]; !ok {
		return d
	}
	return BoolValue(r, k)
}
```

gateway/handlers/reader.go
alexellis: Thanks for looking at this. I thought task states would have been factored into whether a task was ready or not. I asked Eric to try without it and I think we sped things up by a large factor.

Sounds like they might need to come back in then @ericstoekl ?

gateway/handlers/reader.go
alexellis: Didn't pick up on this in the review, good point. Also should this be verbose at the API level (it makes sense at the CLI level) maybe here it should be `inspect` or similar?

gateway/handlers/reader.go
ericstoekl: Yeah, sounds like nodes need to come back. Stefan Prodhan mentioned this as well during the contributors meeting two weeks back. It might be the only way to reliably get replica status.

thaJeztah: I can double check if you want now that SF starts to wake up 😅

ericstoekl: @alexellis There is an issue with having the function filter by `NodeList`, because the client object we have is of type `client.ServiceAPIClient`, which does note implement the `NodeList()` function:

https://github.com/docker/engine-api/blob/4290f40c056686fcaa5c9caf02eac1dde9315adf/client/interface.go#L102

Before [this commit](https://github.com/openfaas/faas/commit/0248a5ea1d9cb412e64df70a13b1337f8a2b625e) `c` was a `client.Client` object, which had the `NodeList()` method available.

alexellis: Hey Eric.. we can work around this by passing in the equivalent interface / struct through from the calling code. You just need the Node API client right? This one? https://github.com/docker/engine-api/blob/4290f40c056686fcaa5c9caf02eac1dde9315adf/client/interface.go#L94

ericstoekl:
Update to reflect suggestions from @alexellis and @thaJeztah. Also added a test for getting replica status.

ericstoekl:
I have synced this up with the latest master branch. How to test:

1. Checkout this PR, build gateway, deploy gateway
2. Run: `curl localhost:8080/system/functions?v=true`
3. You should see a JSON list containing each object. ex:

```
{"name":"func_nodeinfo","image":"functions/nodeinfo:latest@sha256:e6d14cf6bb833fda2cfa7088158efe37b1af1930ef9925094971ad1fd9b7047f","invocationCount":0,"replicas":1,"replicaCount":1,"envProcess":"","labels":{"com.docker.stack.image":"functions/nodeinfo:latest","com.docker.stack.namespace":"func"}},
```

`replicaCount` is included in the output.

I looked at performance, and with the initial set of functions I get ~23ms for the list with `verbose`, and ~18ms for without.

alexellis:
@ericstoekl we'll need to move this over to faas-swarm for merging. Is that something you can work on?

alexellis:
Closing since swarm-specific changes need to target faas-swarm that we're breaking out in #460

PR #330, opened 2017-10-24
[WIP] Enable GET method for functions - proxy fix

### LINKED ISSUE #320: Proposal: Adding mount options to functions.
* Brief summary including motivation/context

The proposed change would allow functions to mount volumes and other directories through the normal docker configuration. This would allow a function to process relatively large amounts of data without having to pass it through http/stdin.

Any design changes

Add docker mount struct to the CreateFunctionRequest struct and passing it along in the  create function handler

Pros + Cons

Pros:

* Functions will be able to have volumes.

Cons:

* Volumes should be available on all nodes running functions (but that is rather expected)

Effort required
Little, it's a two line change.

JockDaRock:
Signed-off-by: Jock Reed jock.da.rock@gmail.com

## Description

Enable passing of the GET method Alex added to enable GET methods.

## Motivation and Context

## How Has This Been Tested?

Tested this in browser and using Postman. Inspected environment variables from flask image as well.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/handlers/proxy.go
+++ b/gateway/handlers/proxy.go
@@ -141,7 +141,8 @@ func invokeService(w http.ResponseWriter, r *http.Request, metrics metrics.Metri
 	contentType := r.Header.Get("Content-Type")
 	fmt.Printf("[%s] Forwarding request [%s] to: %s\n", stamp, contentType, url)
 
-	request, err := http.NewRequest("POST", url, bytes.NewReader(requestBody))
+	method := r.Method
+	request, err := http.NewRequest(method, url, bytes.NewReader(requestBody))
 
 	copyHeaders(&request.Header, &r.Header)
 
--- a/gateway/handlers/proxy.go
+++ b/gateway/handlers/proxy.go
@@ -45,12 +45,13 @@ func MakeProxy(metrics metrics.MetricOptions, wildcard bool, client *client.Clie
 	return func(w http.ResponseWriter, r *http.Request) {
 		defer r.Body.Close()
 
-		if r.Method == "POST" {
+		switch r.Method {
+		case "POST", "GET":
 			logger.Infoln(r.Header)
 
-			xfunctionHeader := r.Header["X-Function"]
-			if len(xfunctionHeader) > 0 {
-				logger.Infoln(xfunctionHeader)
+			xFunctionHeader := r.Header["X-Function"]
+			if len(xFunctionHeader) > 0 {
+				logger.Infoln(xFunctionHeader)
 			}
 
 			// getServiceName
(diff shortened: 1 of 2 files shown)


JockDaRock:
thoughts on this @alexellis?  😄

gateway/handlers/proxy.go
```diff
@@ -140,7 +141,8 @@ func invokeService(w http.ResponseWriter, r *http.Request, metrics metrics.Metri
 	contentType := r.Header.Get("Content-Type")
 	fmt.Printf("[%s] Forwarding request [%s] to: %s\n", stamp, contentType, url)
 
-	request, err := http.NewRequest("POST", url, bytes.NewReader(requestBody))
+	method := r.Method
+	request, err := http.NewRequest(method, url, bytes.NewReader(requestBody))
```
alexellis: We can inline r.Method

alexellis:
How much edge-case testing have you done?

* What happens with a function that expects a body but then doesn't get one?
* Is there a difference between Swarm and Kubernetes (are additional PRs required)?
* Is there a scenario where a GET could be harmful or undesirable?

JockDaRock:
Hmmm... that is interesting.  I don't know why it wouldn't rebase, unless some changes have happened since I posted this. Might have to rework that then.

If a function expects a body and doesn't get one, the function would error out with whatever is built into the function code to handle a no-body request.  This is identical to a POST with no body.  The way I have tested this is enabling POST and GET on a flask route.  The flask function returns the same output when being used on POST and GET.  Functions that are not processing the HTTP  headers will ignore that part of the request anyway.

I have not tried it in my k8s cluster, I would need to update the gateway and give it a try

I can't imagine when GET would be harmful, but maybe I am not being creatvie enough :wink:.  Undesirable... that might be subjective.  To be honest, though, I can imagine a lot of functions that need to run **without** the need for a body, in which case running a GET on a function would be preferable.

Additional thoughts:
There needs to be support for, at the very least, running GET and POST on functions. makes sense for the reason I stated about running functions without a body.  Seems off to do a POST on a function that is not expecting a body, for no other reason than the platforms gateway prevents me from doing any other type of HTTP request.  That and if we are really trying to get a Flask or Express developer to take it seriously, having to put an additional component, like NGINX, in or order to rewrite incoming requests seems a bit too hacky.  That and a dev who is using Flask/Express a website or API function may not have a say on what is deployed in front of it.  which would make it more likely they won't use OpenFaaS.

gateway/handlers/proxy.go
JockDaRock: Yes, that is true, if you would prefer that I can make that change :smile:

JockDaRock:
any other thoughts on the PR @alexellis? 😃

alexellis:
We have a rebasing issue.. so might need to raise this again.

PR #343, opened 2017-10-26
[watchdog] pass request body to process without buffering

### LINKED ISSUE #342: Proposal: watchdog: pass request body to process without buffering
The request body is currently buffered completely in the watchdog. There is no need for that as discussed in #337

## Expected Behaviour
There should be no full buffering of incoming requests in the watchdog

## Current Behaviour
The requests is buffered completely.

## Possible Solution
use io.Reader rather than a byte slice

## Context
I want to be able to process big input data (movies, big archives, etc.)

trusch:
## Description
This change ensures that the request body is passed to the fprocess as a stream rather than a bytes slice.
In most cases this will prevent a complete buffering of the request. The only case where buffering is unfortunatly needed is when marshaling the request in json.

fixes #342

## Motivation and Context
I want to be able to process input data bigger than the available memory of the worker

## How Has This Been Tested?
running the test suite

## Types of changes

## Checklist:

### DIFF
--- a/watchdog/main.go
+++ b/watchdog/main.go
@@ -6,6 +6,7 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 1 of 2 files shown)


alexellis:
Hi @trusch we haven't agreed on the design phase yet, so it's best not to raise PRs for the watchdog work yet. Also this could be a breaking change, any alterations need to be behind a feature flag. Let's talk more on the issue.

Issue #214, opened 2017-09-25
Implement strict RESTful API


imeoer:
## Expected Behaviour

The API definition of gateway and provider should be more in line with the RESTful semantic specification.

> Note: Do not change existing endpoints/routes - so we can maintain compatibility

## Current Behaviour

// List of deployed functions.
`GET /system/functions`
-> 200 Ok

// Deploy a new function.
`POST /system/functions`
-> 200 Ok

// Remove a deployed function.
`DELETE /system/functions`
-> 200 Ok
-> 404 Not Found

// Invoke a deployed function.
`POST /function/$name`
-> 200 Ok
-> 404 Not Found
-> 500 Internal Server Error

// Invoke a deployed function asynchronously 
`POST /async-function/$name`
-> 202 Accepted
-> 404 Not Found
-> 500 Internal Server Error

// Event-sink for AlertManager
`POST /system/alert`
-> 200 Ok
-> 500 Internal Server Error

## Possible Solution

// List of deployed functions.
`GET /system/functions`
-> 200 Ok
-> 500 Internal Server Error

// Deploy a new function.
`POST /system/functions`
-> 201 Created
-> 400 Bad Request
-> 500 Internal Server Error
-> 503 Service Unavailable (FaaS Provider Unreachable)

// Remove a deployed function.
`DELETE /system/functions/$name`
-> 204 No Content
-> 400 Bad Request
-> 404 Not Found
-> 500 Internal Server Error

// Update a deployed function.
`PUT /system/functions/$name`
-> 200 Ok
-> 400 Bad Request
-> 404 Not Found
-> 500 Internal Server Error

// Invoke a deployed function.
`ANY /function/$name`
-> 400 Bad Request
-> 200 Ok
-> 404 Not Found
-> 500 Internal Server Error

// Invoke a deployed function asynchronously 
`ANY /async-function/$name`
-> 400 Bad Request
-> 202 Accepted
-> 404 Not Found
-> 500 Internal Server Error

// Event-sink for AlertManager
`POST /system/alert`
-> 400 Bad Request
-> 200 Ok
-> 500 Internal Server Error

Would be updating gateway, provider, related test cases and Swagger definition.

alexellis:
Can you do a current behaviour vs expected for this please?

imeoer:
@alexellis Updated. where is the `UPDATE` definition in the Swagger doc?

alexellis:
[...]

PR #346, opened 2017-10-26
[WIP] Return 500 if GET /system/functions fails

nullseed:

This is the beginning of the work on creating a more RESTful API (#214). It can be considered for merging on it's own however if there is work already happening to address this issue.

## Description

- If the call to ServiceList fails in the reader, 500 is returned with an error message. Before this change the handler returned 200 with the string `null`.
- If there are no functions then an empty array is returned. Before this change, the string `null` was returned

## Motivation and Context

## How Has This Been Tested?

Running unit tests in `reader_test.go`.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/handlers/reader.go
+++ b/gateway/handlers/reader.go
@@ -6,20 +6,20 @@ package handlers
 import (
 	"context"
 	"encoding/json"
-	"fmt"
+	"log"
 	"net/http"
 
 	"strings"
 
-	"github.com/openfaas/faas/gateway/metrics"
-	"github.com/openfaas/faas/gateway/requests"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
+	"github.com/openfaas/faas/gateway/metrics"
+	"github.com/openfaas/faas/gateway/requests"
 )
 
 // MakeFunctionReader gives a summary of Function structs with Docker service stats overlaid with Prometheus counters.
-func MakeFunctionReader(metricsOptions metrics.MetricOptions, c *client.Client) http.HandlerFunc {
+func MakeFunctionReader(metricsOptions metrics.MetricOptions, c client.ServiceAPIClient) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 
 		serviceFilter := filters.NewArgs()
@@ -30,11 +30,14 @@ func MakeFunctionReader(metricsOptions metrics.MetricOptions, c *client.Client) 
 
 		services, err := c.ServiceList(context.Background(), options)
 		if err != nil {
-			fmt.Println(err)
+			log.Println("Error getting service list:", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte("Error getting service list"))
+			return
 		}
 
 		// TODO: Filter only "faas" functions (via metadata?)
-		var functions []requests.Function
+		functions := []requests.Function{}
 
 		for _, service := range services {
 
--- a/gateway/tests/reader_test.go
+++ b/gateway/tests/reader_test.go
@@ -1,0 +1,208 @@
+package tests
+
+import (
+	"encoding/json"
+	"errors"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/swarm"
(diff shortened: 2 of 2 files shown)


gateway/handlers/reader.go
```diff
@@ -6,20 +6,20 @@ package handlers
...
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
+	"github.com/openfaas/faas/gateway/metrics"
+	"github.com/openfaas/faas/gateway/requests"
 )
 
 // MakeFunctionReader gives a summary of Function structs with Docker service stats overlaid with Prometheus counters.
-func MakeFunctionReader(metricsOptions metrics.MetricOptions, c *client.Client) http.HandlerFunc {
+func MakeFunctionReader(metricsOptions metrics.MetricOptions, c client.ServiceAPIClient) http.HandlerFunc {
```
alexellis: What changed RE: client.ServiceAPIClient?

gateway/handlers/reader.go
```diff
@@ -30,11 +30,14 @@ func MakeFunctionReader(metricsOptions metrics.MetricOptions, c *client.Client)
...
-			fmt.Println(err)
+			log.Println("Error getting service list:", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte("Error getting service list"))
+			return
 		}
 
 		// TODO: Filter only "faas" functions (via metadata?)
-		var functions []requests.Function
+		functions := []requests.Function{}
```
alexellis: This changed?

gateway/handlers/reader.go
nullseed: Using `client.ServiceAPIClient` allowed me to mock the call to `c.ServiceList` in the unit tests.

gateway/handlers/reader.go
nullseed: It did! When functions was `nil` (there were no services in the swarm) then `null` was returned in the body of the HTTP response. Now if there are no services, an empty array `[]` is returned instead.

gateway/handlers/reader.go
alexellis: Is that an interface in the Docker code?

gateway/handlers/reader.go
alexellis: OK makes sense.

gateway/tests/reader_test.go
```diff
@@ -0,0 +1,208 @@
...
+	return nil, nil
+}
+
+func (c testServiceApiClient) TaskInspectWithRaw(ctx context.Context, taskID string) (swarm.Task, []byte, error) {
+	return swarm.Task{}, []byte{}, nil
+}
+
+func (c testServiceApiClient) TaskList(ctx context.Context, options types.TaskListOptions) ([]swarm.Task, error) {
+	return []swarm.Task{}, nil
+}
```
alexellis: It's kind of a shame to have so many methods here including Remove/Create etc. Might be necessary for using the interface verbatim. the adapter pattern could allow for a smaller interface implementation.

gateway/handlers/reader.go
nullseed: Yep https://github.com/openfaas/faas/blob/master/gateway/vendor/github.com/docker/docker/client/interface.go#L124

gateway/tests/reader_test.go
nullseed: That's true. I wanted to avoid changing `server.go` so I went with the approach of using the interface from docker. Do you want me to create an adapter for this PR or address it in the following REST-related PRs?

alexellis: I'll merge this as-is, but yes It think we should be careful of having this sort of repetition.

alexellis:
Thanks @nullseed - can you pull in master on your machine and rebuild this? I'd like to see this tested with master before releasing. If you can work with someone else from Slack that'd be even better.

nullseed:
@alexellis tested on master, no regressions that I can see. One thing i did notice was that after the reader handler is wrapped in the metrics handler, headers such as Content-Type are not forwarded. That means that GET /system/functions always returns `text/html` even if the reader returns `application/json`.

alexellis:
> One thing i did notice was that after the reader handler is wrapped in the metrics handler

Could you identify the line and suggest a fix for this?

nullseed:
@alexellis see #351

PR #348, opened 2017-10-27
Add Http_ContentLength for deterministic functions

alexellis:

## Description

Add Http_ContentLength for deterministic functions

## Motivation and Context

Enables deterministic reading from stdin. cc/ @Lewiscowles1986 

## How Has This Been Tested?

With unit tests and the `env` built-in shell command.

Set process to `env` and pass a body of `test`

```
...
Http_ContentLength=4
...
```

See README.md for updates

## Types of changes

## Checklist:

### DIFF
--- a/watchdog/main.go
+++ b/watchdog/main.go
@@ -202,6 +202,7 @@ func getAdditionalEnvs(config *WatchdogConfig, r *http.Request, method string) [
 		}
 
 		envs = append(envs, fmt.Sprintf("Http_Method=%s", method))
+		envs = append(envs, fmt.Sprintf("Http_ContentLength=%d", r.ContentLength))
 
 		if config.writeDebug {
 			log.Println("Query ", r.URL.RawQuery)
--- a/watchdog/README.md
+++ b/watchdog/README.md
@@ -91,7 +91,8 @@ Headers and other request information are injected into environmental variables 
 The `X-Forwarded-By` header becomes available as `Http_X_Forwarded_By`
 
 * `Http_Method` - GET/POST etc
-* `Http_Query` - Querystring value
+* `Http_Query` - QueryString value
+* `Http_ContentLength` - gives the total content-length of the incoming HTTP request received by the watchdog.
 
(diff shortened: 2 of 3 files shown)


watchdog/main.go
```diff
@@ -202,6 +202,7 @@ func getAdditionalEnvs(config *WatchdogConfig, r *http.Request, method string) [
 		}
 
 		envs = append(envs, fmt.Sprintf("Http_Method=%s", method))
+		envs = append(envs, fmt.Sprintf("Http_ContentLength=%d", r.ContentLength))
```
Lewiscowles1986: Minor gripe, although I've already used this name. How important should the transport method be to the code dealing with this?

alexellis: What is the gripe, can you explain?

Lewiscowles1986:
Also does this need someone to work on `faas-cli` or is someone else handling?

alexellis:
@Lewiscowles1986 good input. I believe this data is passed by the HTTP client automatically.

watchdog/main.go
Lewiscowles1986: Sorry never noticed this. What I was saying was should the variable care and convey that this information is passed via Http?

It doesn't matter anyway, I've added it to the function I wanted to call. Functionality being more important than abstract correctness and all that :wink: :+1:

alexellis: This is how data is passed via CGI to running applications.

https://users.cs.cf.ac.uk/Dave.Marshall/PERL/node200.html

http://www.cgi101.com/book/ch3/text.html

We have just added a namespace to avoid collisions.

Lewiscowles1986: the point was specifically about the prefix. The linked CGI example just calls it content-length. Like I say it's no big deal at all. It's more of a big-deal for the code I want to include it in because it shouldn't know or care about HTTP. However I could have instead wrapped it in a shell script so I guess it's as much on my use of the ENV as the name.

Issue #349, opened 2017-10-27
Research: always-on/generated token for UI and API


alexellis:
It's very hard to automate security out of the box for an API/UI because this generally involves:

* non-default username/password
* tokens
* CA/certs

We should force always-on/generated auth for the UI and API via basic auth or a token, so that if someone does deploy in production (without reading the best practice guide) they are not wide open to vulnerabilities.

alexellis:
Ideas? @chanezon @chanwit @Lewiscowles1986 + Joe Beda.

This would be at the API gateway level - so agnostic to the scheduler so it can work with Swarm, K8s and other back-ends.

We could have a user create a secret or token (using the orchestrator) ahead of time then fetch that back before accessing the cluster.

Lewiscowles1986:
I can generate a secure'ish random string and sed it in as part of setup. I can also generate two for basic-auth and feel I have a good handle on wrangling basic-auth into the gateway services with a middleware.

For the record I don't like the idea. What happens if someone insists they need kerberos next?

Perhaps we could instead add to tutorials a kong setup. I've never used Kong and would enjoy the opportunity to play, get feedback which feeds value in both directions. That way it's not part of FaaS, but is part of stock scripts.

alexellis:
It needs to be a one click deployment. We like Kong but it needs a multi stage manual setup. Suggestions welcome

alexellis:
Derek close: fixed with basic auth in GW

PR #350, opened 2017-10-28
Suggestion - basic auth security with Nginx

alexellis:
## Description

Minimal example for validation / testing to setup basic auth with Nginx for the whole gateway.

## Motivation and Context

Context - #349

This requires some manual steps which means it won't be possible to do a `docker stack deploy` and get going. Here's the workflow:

* Create .htpassword using CLI
* Create Docker secret using CLI
* Deploy stack

The main change is that the gateway would not be exposed, but Nginx would be instead.

### DIFF
--- a/contrib/nginx/.gitignore
+++ b/contrib/nginx/.gitignore
@@ -1,0 +1,1 @@
+openfaas.htpasswd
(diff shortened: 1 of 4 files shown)


rorpage:
In the `README.md` you say: "Build gwnginx from contrib directory."

I propose adding the build command for creating that Docker image.

`docker build -t gwnginx .`

I just tested this using the instructions and it worked perfectly! :+1:

alexellis:
Thanks @rorpage - can you try adding a section in the conf file that allows /function to be accessed publicly but prevents everything else? I think generally functions will want to implement their own checking and right now it requires event sources like Alexa or GitHub to "speak basic auth".

PR #351, opened 2017-10-28
Return correct Content-Type from an AddMetricsHandler wrapped handler

nullseed:

From testing master after #346 was merged: after the reader handler is wrapped in the metrics handler, headers such as Content-Type are not forwarded. That means that GET /system/functions always returns text/html even if the reader returns application/json. This can be observed in postman by issuing the above request. This is existing behaviour.

This PR addresses the issue above and adds unit tests that describe the changes.

Note that add_metrics_test.go is in the metrics package so that coverage of the add_metrics.go statements could be seen.

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/gateway/metrics/add_metrics_test.go
+++ b/gateway/metrics/add_metrics_test.go
@@ -1,0 +1,81 @@
+package metrics
+
+import (
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+type testFetchQuery struct {
+	res *VectorQueryResponse
+	err error
+}
+
+func (f testFetchQuery) Fetch(query string) (*VectorQueryResponse, error) {
+	return f.res, f.err
+}
+
+func TestAddMetricsHandlerServeHTTPReturnsContentTypeOfPassedInHandlerWhenFetchSucceeds(t *testing.T) {
+	contentTypes := []string{
+		"text/html",
+		"application/json",
+	}
+
+	for _, contentType := range contentTypes {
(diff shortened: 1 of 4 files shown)


alexellis:
Alex you're certainly making your mark here. Thank you for the attention to detail.

gateway/metrics/add_metrics_test.go
```diff
@@ -0,0 +1,81 @@
...
+type testFetchQuery struct {
+	res *VectorQueryResponse
+	err error
+}
+
+func (f testFetchQuery) Fetch(query string) (*VectorQueryResponse, error) {
+	return f.res, f.err
+}
+
+func TestAddMetricsHandlerServeHTTPReturnsContentTypeOfPassedInHandlerWhenFetchSucceeds(t *testing.T) {
```
alexellis: I am not clear what these tests are doing. We always need to return Content-Type of `application/json`.

nullseed: These tests prove the issue described in the PR. They can be extended to cover more of the logic in the AddMetricsHandler function.

alexellis: I think the function needs to return `application/json` the whole time, unless it's an error as in an exception.

alexellis:
I think one of the Pos that went in recently addressed the "null" response. Would you like to verify whether this is still necessary?

alexellis:
Closing as I'm not sure if this is necessary anymore. Have pinged 4 days ago.

PR #352, opened 2017-10-28
[Proposal] Move test directly into package and add test verbosity

nullseed:

To see coverage for hander tests I've moved reader_test.go into the handlers package. Now instead of the build showing:

```
?   	github.com/openfaas/faas/gateway/handlers	[no test files]
```

it shows:

```
ok  	github.com/openfaas/faas/gateway/handlers	0.023s	coverage: 5.3% of statements
```

I've also added the `-v` flag to the go test command to show tests that have passed/failed like in the watchdog build.

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/gateway/handlers/reader_test.go
+++ b/gateway/handlers/reader_test.go
@@ -1,4 +1,4 @@
-package tests
+package handlers
 
 import (
 	"encoding/json"
@@ -10,7 +10,6 @@ import (
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/swarm"
-	"github.com/openfaas/faas/gateway/handlers"
 	"github.com/openfaas/faas/gateway/metrics"
 	"github.com/openfaas/faas/gateway/requests"
 	"golang.org/x/net/context"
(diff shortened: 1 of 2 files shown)


alexellis:
Derek add label: proposal

gateway/handlers/reader_test.go
```diff
@@ -1,4 +1,4 @@
-package tests
+package handlers
```
alexellis: I'm not sure that I like the style of putting tests inside the Go packages.

nullseed: Ok. It allows you to view how many statements are covered in the package. There may be another way to do this however. It's also convention in a lot of go projects. See https://golang.org/src/net/http/ and https://github.com/moby/moby/tree/master/client

alexellis: We should still propose changes that are counter to the existing conventions in the project. If that's something we decide to do we'd need to do it across the board.

nullseed: Yep

alexellis:
I can see a merge conflict. If we decide to do this as a group then we'll need to do it across the board. Closing for now. Feel free to keep commenting and raise a proposal for future design changes.

Issue #414, opened 2017-11-29
Feature: Add min/max scale for auto-scaling - Swarm & external providers


alexellis:
Should be able to set a minimum scale for functions as well as the maximum (existing)

We can configure this through labels on the Docker Swarm service.

```
  labels:
    "com.openfaas.scale.min": "5"
    "com.openfaas.scale.max": "15"
```

Labels are generic so this would need faas-netes to expose some new data for the /system/function/<name> call too.

This will be useful for functions which are known to need to have high availability or traffic.

Implementation on Swarm:
* [x] Alert handler
* [x] Create handler
* [x] Update handler

cc @stefanprodan @ericstoekl 

Initial work: https://github.com/openfaas/faas/compare/scale_with_min_values?expand=1

stefanprodan:
Tested on Kubernetes with the following stack:

```yml
functions:
  nodeinfo:
    handler: node main.js
    image: functions/nodeinfo:burner
    labels:
      com.openfaas.scale.min: "5"
      com.openfaas.scale.max: "15"
```

It got up to 15 PODs during the load test and down to 5 after that:

<img width="1026" alt="screen shot 2017-11-30 at 00 14 48" src="https://user-images.githubusercontent.com/3797675/33401898-980aff0e-d563-11e7-9c6d-8d058a93781d.png">

alexellis:
Thank you for helping to test this on Kubernetes Stefan.

That result sounds like what we want. When the settings are not present does it fallback to the old behavior?

stefanprodan:
With no `com.openfaas.scale.min` label it goes back to one pod.

alexellis:
Derek add label: enhancement

austinfrey:
Is this closeable with the merge of #419?

alexellis:
Derek close

alexellis:
Yes thanks for helping out

white-cloud:
Hello please how i can autoscale my faas fucntion ??

white-cloud:
which command i use for example to fix the number of replica depend on the CPU usage or the number of requests ??

alexellis:
@white-cloud this thread is closed - please open a new issue if you'd like support.

alexellis:
Derek lock: feature completed

PR #419, opened 2017-12-02
Feature: Add min/max scale for auto-scaling - Swarm & external providers 

alexellis:

## Description

Feature: Add min/max scale for auto-scaling - Swarm & external providers 

## Motivation and Context

See: #414 

## How Has This Been Tested?

With new tests and with Stefan via gateway and kubernetes

## Types of changes

## Checklist:

### DIFF
--- a/TestDrive.md
+++ b/TestDrive.md
@@ -242,7 +242,7 @@ The `-d` value passes in the argument for your function. This is read via STDIN 
 
 Grab OS, CPU and other info via a Node.js container using the `os` module.
 
-If you invoke this method in a while loop or with a load-generator tool then it will auto-scale to 5, 10, 15 and finally 20 replicas due to the load. You will then be able to see the various Docker containers responding with a different Hostname for each request as the work is distributed evenly.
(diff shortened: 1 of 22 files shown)


ericstoekl:
Tried this out and at least can see that `min` is working, as the number of replicas for my function starts at 5. What's the recommended way to make the container scale up until it hits its `max` limit? Would using `ab` work?

alexellis:
Yes that will work too with a high number or a bash loop. Doesn't need to be high figures for concurrency and you can monitor the alerts in Prometheus too

ericstoekl:
Gotcha, I will try that out.

I have tried to load this up on Kubernetes (uploaded the gateway image to docker hub, then set the gateway definition in `faas.yml` to load that image), and it looks like my function has the min=5 and max=15 labels set, but it only has one replica when deployed. Would this be expected to work on faas-netes, or would more work be needed first?

stefanprodan:
@ericstoekl this how Kubernetes works, replicas refer to pods not containers

alexellis:
@ericstoekl  do this instead - `kubectl get deploy` that is where the replica count is stored.

ericstoekl:
In that case, I see that the function has the correct number of replicas. Thanks 💯

alexellis:
Thank you both for testing. I'll merge.

Issue #312, opened 2017-10-18
Function return Content-Type not honored by Gateway


mcgoober:
## Expected Behaviour
The watchdog has a setting 'content_type' which when set forces the return content-type of a function. This content-type that is returned by the watchdog should be also be returned by the gateway.

## Current Behaviour
The gateway returns the same content-type as the request. i.e if you POST with content-type 'application/json' in the request, then 'application/json' is in the response, no matter what the function actually returns.

## Possible Solution
https://github.com/openfaas/faas/blob/master/gateway/handlers/proxy.go#L173
At the time of writing...
https://github.com/openfaas/faas/blob/dde98eb582c381d07d4b7c6842b815c455179b57/gateway/handlers/proxy.go#L173

This line of code forces response content-type to the request and overwrites the function return content-type. The comment above the line suggests that this may been thought about.

Removal of this line prevents the content-type from being overwritten and allows the gateway to return the correct content-type for the function.

## Steps to Reproduce (for bugs)
1. Create a function with env var 'content_type' set to 'text/html'
2. POST to the function with request content-type say.. 'application/json'
3. Function will return 'application/json', this is true for any content-type

## Context
Found this issue whilst attempting to make a function that returned HTML for a browser and thus develop a web application as a function or set of functions. The browser won't display the returned HTML  unless the content-type is 'text/html'.  
This solution also requires the ability to call a function with a GET, but I've accomplished this in another way.

## Your Environment
* Gateway 0.6.6d
* Watchdog 0.6.6d
* Docker Swarm 17.06.1-ce

alexellis:
Thanks for using OpenFaaS @mcgoober 

I am sure I have this working correctly in my recent Alexa skills - but it may be as you say because of the caller. I'll look into this over the next few days. We can also take a PR if you want to move quicker than that.

You can also checkout the new code branch I pushed on for enabling "GET" routes. We published some info about it in the Slack channel #general I believe.

alexellis:
Thank you for picking this up @ericstoekl

alexellis:
@ericstoekl I had some time to work on this on the flight.

alexellis:
I've implemented the new behaviour with a fall-back to old behaviour when the variable is not defined with a value.

mcgoober:
Thanks @alexellis for the fix. However the behavior hasn't changed in 0.6.7. 
The pull request from this ticket would have worked as far as I can see...
[...]

PR #435, opened 2017-12-13
Remove Content-Type forwarding from Request

ericstoekl:
Signed-off-by: Eric Stoekl <ems5311@gmail.com>

This PR removes forwarding of Request Content-Type to the gateway.

## Description

Changes the code to only consider Response Content-Type, and if Response is empty, use the default value. No longer consider Request Content-Type.

## Motivation and Context

https://github.com/openfaas/faas/issues/312#issuecomment-351067243

From #312 

## How Has This Been Tested?

Tests have been updated to reflect this change, which was mostly removing code that tested for Request Content-Type being forwarded.

Also tested manually using the repro steps detailed here (https://github.com/openfaas/faas/issues/312#issuecomment-351030182)

## Types of changes

## Checklist:


gateway/handlers/proxy.go
```diff
@@ -162,7 +162,7 @@ func invokeService(w http.ResponseWriter, r *http.Request, metrics metrics.Metri
 
 	defaultHeader := "text/plain"
 
-	w.Header().Set("Content-Type", GetContentType(response.Header, r.Header, defaultHeader))
+	w.Header().Set("Content-Type", GetContentType(response.Header, defaultHeader))
```
alexellis: Do we need this line anymore?

ericstoekl: I think it is necessary because `w.Header` needs to be set with the proper ContentType. If we don't do this, then the header won't be set with either the response header or the default.

Also, `clientHeader` defined above at line 160 appears to be created but not used:
https://github.com/openfaas/faas/pull/435/files/f1c825003d86b55242cb9e55e250d03fe4784ceb#diff-a7739e9c4834f2d0769169a1c27e7774R160

ericstoekl:
Pushed an update removing the `GetContentType()` method and the tests for it as it seems it is no longer necessary.

alexellis:
Looks fine to me, it will be a breaking change to behaviour so we'll need to test it properly.

The first thing that will break is any Alexa skill that people develop - it's part of the reason this behaviour was there in the first place. cc @rgee0

Issue #519, opened 2018-02-16
UI: update portal to work over `kubectl proxy`


alexellis:
## Expected Behaviour

If possible the portal UI should work over kubectl proxy

## Current Behaviour

This fails with an error due to the browser trying to query the root of the domain. The kubectl proxy forces a deep relative path so the XHR request fails.

## Possible Solution

Make a non-root call to the /system/functions endpoint? Not sure what the best approach is here.

cc @kenfdev @jrevillas 

## Steps to Reproduce (for bugs)

1. Deploy kubernetes and OpenFaaS
2. `kubectl proxy`
3. Head to 127.0.0.1:8001/api/v1/proxy/namespaces/openfaas/services/gateway/

## Context

Mentioned in following issue: https://github.com/openfaas/faas-netes/issues/63

## Your Environment

* Docker version `docker version` (e.g. Docker 17.0.05 ):

* Are you using Docker Swarm or Kubernetes (FaaS-netes)?
DfM

* Operating System and version (e.g. Linux, Windows, MacOS):
MacOS

* Please also follow the [troubleshooting guide](https://github.com/openfaas/faas/blob/master/guide/troubleshooting.md) and paste in any other diagnostic information you have:

alexellis:
Reported by @id0Sch originally

kenfdev:
I assume we can fix this by using relative paths for the ajax calls.
I think traefik is doing it the same way looking at this [PR](https://github.com/containous/traefik/pull/492).

I can look at this in the following week.

alexellis:
Thanks @kenfdev

alexellis:
Merged. Let's close the issue when we have confirmed the fix in 0.7.2

alexellis:
<img width="1045" alt="screen shot 2018-02-25 at 13 08 10" src="https://user-images.githubusercontent.com/6358735/36641741-f4c609c4-1a2c-11e8-8430-eb7085c9abdb.png">

Did I mention it yet? @kenfdev is awesome

alexellis:
Derek close: fixed

PR #529, opened 2018-02-22
Fixed UI to use relative paths for ajax requests in order to work with `kube proxy`

kenfdev:
Signed-off-by: Ken Fukuyama <fukuyama@supersoftware.co.jp>

## Description

Because the ajax requests used the paths from the root, the gateway UI didn't work correctly via `kube proxy`. Using relative paths for the requests would solve this problem.
This is the [same fix](https://github.com/containous/traefik/pull/492) traefik has made, too.

## Motivation and Context

People who use the gateway via `kube proxy` cannot use the UI currently.
This fixes issue #519.

## How Has This Been Tested?

I've spun up a kubernetes cluster with docker on mac. Deployed openfaas and checked

* Create
* Delete
* Invoke (downloaded an image, too)

both with and without `kube proxy`.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/assets/script/bootstrap.js
+++ b/gateway/assets/script/bootstrap.js
@@ -53,7 +53,7 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$timeout', '$md
             }
 
             var options = {
-                url: "/function/" + $scope.selectedFunction.name,
+                url: "../function/" + $scope.selectedFunction.name,
                 data: $scope.invocation.request,
                 method: "POST",
                 headers: { "Content-Type": requestContentType },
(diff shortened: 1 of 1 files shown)


gateway/assets/script/bootstrap.js
```diff
@@ -143,7 +143,7 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$timeout', '$md
             var previous = $scope.functions;
 
             var cl = function(previousItems) {
-                $http.get("/system/functions").then(function(response) {
+                $http.get("../system/functions").then(function(response) {
```
alexellis: is there any likelihood that nginx, caddy or some other ingress could block a relative path like "../"? cc @stefanprodan

stefanprodan: It should be fine, if it works with kube proxy it should also work with nginx sub paths

alexellis: Thanks for your advice. I'll merge.

alexellis:
Will release 0.7.2 after the Angular upgrade in #531

Issue #538, opened 2018-02-25
Watchdog: marshal "Cookie" header into environment


alexellis:
## Description

`Http_Cookie` should contain the value passed in the `Cookie` header if it is not already present.

Area for change:

https://github.com/openfaas/faas/blob/master/watchdog/main.go#L206

This change should also cover https://github.com/openfaas-incubator/of-watchdog

Work items:

* First of all - test that the the Cookie is not already transmitted by forcing a header at the gateway

`curl -H "Cookie: some-value" etc....`

* Update the line mentioned in the watchdog and test with setting the `fprocess` env-var to `env` (you can run this on your host without a container for testing

* Next update the of-watchdog with the same change and test in the same way

alexellis:
Suggested contributors: @johnmccabe @dungeonmaster18 @viveksyngh

alexellis:
Derek add label: area/watchdog

viveksyngh:
I will work on this.

johnmccabe:
Just sanity checked this and it appears to be unnecessary. The `Cookie` header, if present, is already resulting in `Http_Cookie` being set in the function env.

https://github.com/openfaas/faas/blob/8f8e337dccb66b3e1dcdc9287e0b037a9937c332/watchdog/main.go#L208-L211

johnmccabe:
Derek close: already implemented

PR #539, opened 2018-02-25
Populate `Http_Cookie` env-var if Cookie header is set.

johnmccabe:
## Description
Populate Http_cookie envvar if Cookie header is set.

Note that this ignores all but the first Cookie header, per [rfc6265#section-5.4]( https://tools.ietf.org/html/rfc6265#section-5.4) which stipulates:

> When the user agent generates an HTTP request, the user agent MUST
> NOT attach more than one Cookie header field.

## Motivation and Context

- [x] I have raised an issue to propose this change (#538)

## How Has This Been Tested?

Built function using updated watchdog invoking env.

image

## Types of changes

## Checklist:

### DIFF
--- a/watchdog/main.go
+++ b/watchdog/main.go
@@ -203,6 +203,11 @@ func pipeRequest(config *WatchdogConfig, w http.ResponseWriter, r *http.Request,
 func getAdditionalEnvs(config *WatchdogConfig, r *http.Request, method string) []string {
 	var envs []string
 
+	cookies := r.Cookies()
+	if len(cookies) != 0 {
+		envs = append(envs, fmt.Sprintf("Http_Cookies=%s", cookies[0].String()))
+	}
+
 	if config.cgiHeaders {
 		envs = os.Environ()
 		for k, v := range r.Header {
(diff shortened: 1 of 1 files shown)


watchdog/main.go
```diff
@@ -203,6 +203,11 @@ func pipeRequest(config *WatchdogConfig, w http.ResponseWriter, r *http.Request,
 func getAdditionalEnvs(config *WatchdogConfig, r *http.Request, method string) []string {
 	var envs []string
 
+	cookies := r.Cookies()
+	if len(cookies) != 0 {
+		envs = append(envs, fmt.Sprintf("Http_Cookies=%s", cookies[0].String()))
+	}
```
alexellis: I think this should be `Http_Cookie` and placed inside the if-guard on `config.cgiHeaders`.

johnmccabe:
Just checked, this isn't actually needed. The header *does* get populated already from `r.Header`.

@alexellis was there some some of special case you'd been thinking of here.

Closing this PR as invalid.

alexellis:
Understood.. that's... Convenient. Sorry if anyone did any unnecessary work.

PR #556, opened 2018-03-03
Make UI block "Invoke" button until function reports readiness

### LINKED ISSUE #555: Feature: Add readiness to the functions in the UI
## Expected Behaviour

We should see if a function is ready to call.

* Invoke should be disabled if not
* A new field should indicate if it's ready

## Current Behaviour

* Error given - Service unavailable

## Possible Solution

Poll our new endpoint that exposes function readiness in the `availableReplicas` field.

`/system/function/<name>` 

## Impact

* faas-netes, faas-swarm and faas-nomad

## Mock-ups

querying

ready

alexellis:

## Description

Fixes #555 

## Motivation and Context

## How Has This Been Tested?

Tested with faas-netes which provides `availableReplicas` on the `/system/function/<name>` endpoint.

This change requires a change to faas-swarm and faas-nomad to report `availableReplicas` = 1 all the time, or optionally when the underlying container is ready.

Update the gateway to the `functions/gateway:pr554` for testing

## Types of changes

Breaking change to UI only, not to the API.

## Checklist:

### DIFF
--- a/gateway/assets/script/bootstrap.js
+++ b/gateway/assets/script/bootstrap.js
@@ -36,24 +36,23 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$timeout', '$md
         }
 
         $scope.invocation.request = "";
-        $scope.isFunctionReady = "Querying";
-        $scope.fetchFunctionsDelay = 3500;
-        $scope.queryFunctionDelay = 2500;
+        var fetchFunctionsDelay = 3500;
+        var queryFunctionDelay = 2500;
         
-        $scope.fetchFunctionsInterval = setInterval(function() {
+        var fetchFunctionsInterval = setInterval(function() {
             refreshData();
-        }, $scope.fetchFunctionsDelay);
+        }, fetchFunctionsDelay);
 
-        $scope.queryFunctionInterval = setInterval(function() {
+        var queryFunctionInterval = setInterval(function() {
             if($scope.selectedFunction && $scope.selectedFunction.name) {
                 refreshFunction($scope.selectedFunction);
             }
-        }, $scope.queryFunctionDelay);
+        }, queryFunctionDelay);
 
         var refreshFunction = function(functionInstance) {
             $http.get("../system/function/" + functionInstance.name)
             .then(function(response) {
-                functionInstance.ready = (response.data && response.data.availableReplicas && response.data.availableReplicas >0);
+                functionInstance.ready = (response.data && response.data.availableReplicas && response.data.availableReplicas > 0);
             })
             .catch(function(err) {
                 console.error(err);
--- a/gateway/assets/script/bootstrap.js
+++ b/gateway/assets/script/bootstrap.js
@@ -31,11 +31,37 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$timeout', '$md
             labels: {}
         };
 
+        $scope.isReady = function(selectedFunction) {
+            if(selectedFunction.ready != undefined && selectedFunction.ready) {
+                return "Ready";
+            }
+            return "Querying..";
+        }
+
         $scope.invocation.request = "";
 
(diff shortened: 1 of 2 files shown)


alexellis:
@kenfdev please could you review / test? @dungeonmaster18 could you help with testing?

dungeonmaster18:
Yup I will do it in the morning tomorrow.

gateway/assets/script/bootstrap.js
```diff
@@ -4,6 +4,16 @@
 
 var app = angular.module('faasGateway', ['ngMaterial', 'faasGateway.funcStore']);
 
+app.directive("readyInput", function() {
```
kenfdev: We should avoid using `directive`s if we aren't dynamically manipulating the DOM (e.g. adding/removing with `jqlite`) . We should use `component` instead and isolate the `scope` to reduce shared state.

In this case, it might be good enough to just add the `md-input-container` in the existing `html` without creating another directive (component), imo.

gateway/assets/script/bootstrap.js
```diff
@@ -31,11 +41,34 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$timeout', '$md
...
+        $scope.fetchFunctionsDelay = 3500;
+        $scope.queryFunctionDelay = 2500;
+        
+        $scope.fetchFunctionsInterval = setInterval(function() {
             refreshData();
-        }, 1000);
+        }, $scope.fetchFunctionsDelay);
+
+        $scope.queryFunctionInterval = setInterval(function() {
+            if($scope.selectedFunction && $scope.selectedFunction.name) {
```
kenfdev: I have to see the details but this statement seems strange...

`$scope.selectedFunction && $scope.selectedFunction.name`

It looks like `selectedFunction` has multiple types (sometimes it's an `object` , and sometimes it's `string` ?)

gateway/assets/script/bootstrap.js
alexellis: I think it could be inlined again now. I had issues with ng-model now allowing a method to be called.

gateway/assets/script/bootstrap.js
kenfdev: my bad... I mislooked and though it was

```
$scope.selectedFunction === $scope.selectedFunction.name
```

ignore this comment please

gateway/assets/script/bootstrap.js
kenfdev: hmm... shared `$scope` is an anti-pattern in AngularJS. I can add an example tonight with a `component`.  

> I had issues with ng-model now allowing a method to be called

I would like to hear what happened here. Would like to reproduce the problem.

I'm going out today so it'll be at least 12 hours from now 😉

kenfdev:
Not sure if I did it the right way but added a PR ( #558 ).


About the new feature:
It would be better if we could retrieve the `availableReplicas` when requesting `system/functions`. Because every time we refresh the UI, it starts with a "Not Ready" state and clicking the function the first time always leads to a Disabled Invoke button. (I guess this isn't possible with the API now)
But it's better than invoking a not ready function imo.
Nice feature!

alexellis:
Thanks for the PR to this remote branch. 

This new field could be populated in `system/functions` for Kubernetes at no cost, but for Docker Swarm this check introduces a significant overhead that could run into several seconds depending on the specs of the nodes and how many there are.

In the interests of improving the user-experience I am OK with the brief delay when clicking between functions. It would be ideal if we could populate the information at the `/system/functions` level without a penalty for Swarm.

kenfdev:
The UI feature LGTM :)

Tested with:
* Chrome
* Safari
* Firefox

Functions:
* Inception(json)
* Figlet(text)
* QRCode(download)

Test Cases:
create/invoke/delete

PR #560, opened 2018-03-04
fixed unexpected behavior where the selected function properties don't get updated

kenfdev:
Signed-off-by: Ken Fukuyama <kenfdev@gmail.com>

## Description

Suppose you had function A and B.
Delete function B and then select function A before B gets deleted from the UI.

When function B disappears from the UI, function A's Status becomes `Not Ready`.
This is because, when the `$scope.functions` updates its property by fetching from `system/functions`, the `$scope.selectedFunction` object changes (even though all the properties are the same, it's a different object.

This means

```
refreshFunction($scope.selectedFunction);
```

isn't updating the newly fetched version of the `selectedFunction`, hence even though it's `Ready`, the UI doesn't get updated.

There may be a better way but for now what I came up of is to update the selected function when the `$scope.functions` length changes.

Plus, I found some anti-patterns and fixed them as well. (if not appropriate, I'll create a different issue and PR)

## How Has This Been Tested?

Chrome

The instructions are explained in the Description.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/assets/index.html
+++ b/gateway/assets/index.html
@@ -72,7 +72,7 @@
                 <div flex></div>
             </md-content>
 
-            <md-content flex layout="column" ng-repeat="function in functions" ng-show="function.name == selectedFunction.name">
+            <md-content flex layout="column" ng-repeat="function in functions" ng-if="function.name == selectedFunction.name">
 
                 <md-card md-theme="default" md-theme-watch>
                     <md-card-title>
--- a/gateway/assets/script/bootstrap.js
+++ b/gateway/assets/script/bootstrap.js
@@ -4,8 +4,8 @@
 
 var app = angular.module('faasGateway', ['ngMaterial', 'faasGateway.funcStore']);
 
-app.controller("home", ['$scope', '$log', '$http', '$location', '$timeout', '$mdDialog', '$mdToast', '$mdSidenav',
-    function($scope, $log, $http, $location, $timeout, $mdDialog, $mdToast, $mdSidenav) {
+app.controller("home", ['$scope', '$log', '$http', '$location', '$interval', '$filter', '$mdDialog', '$mdToast', '$mdSidenav',
+    function($scope, $log, $http, $location, $interval, $filter, $mdDialog, $mdToast, $mdSidenav) {
         var newFuncTabIdx = 0;
         $scope.functions = [];
         $scope.invocationInProgress = false;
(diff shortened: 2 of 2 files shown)


gateway/assets/index.html
```diff
@@ -72,7 +72,7 @@ <h1 style="flex: 1 1 auto;">&nbsp; OpenFaaS Portal</h1>
                 <div flex></div>
             </md-content>
 
-            <md-content flex layout="column" ng-repeat="function in functions" ng-show="function.name == selectedFunction.name">
```
kenfdev: `ng-show` creates DOM elements and is just hiding it with `css`. If we have tens of functions, the tens of DOM will be created event though only 1 is showing.

gateway/assets/script/bootstrap.js
```diff
@@ -31,19 +31,15 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$timeout', '$md
...
-        $scope.isReady = function(selectedFunction) {
-            return (selectedFunction.ready != undefined && selectedFunction.ready == true);
-        }
-
         $scope.invocation.request = "";
         var fetchFunctionsDelay = 3500;
         var queryFunctionDelay = 2500;
         
-        var fetchFunctionsInterval = setInterval(function() {
+        var fetchFunctionsInterval = $interval(function() {
```
kenfdev: `setInterval` isn't aware of the AngularJS context so is better to use the `$interval` service prepared by AngularJS.

gateway/assets/script/bootstrap.js
```diff
@@ -169,6 +165,14 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$timeout', '$md
...
                         if (previousItems.length != response.data.length) {
                             $scope.functions = response.data;
+                            
+                            // update the selected function object because the newly fetched object from the API becomes a different object
+                            var filteredSelectedFunction = $filter('filter')($scope.functions, {name: $scope.selectedFunction.name}, true);
+                            if (filteredSelectedFunction && filteredSelectedFunction.length > 0) {
+                                $scope.selectedFunction = filteredSelectedFunction[0];
+                            } else {
+                                $scope.selectedFunction = undefined;
+                            }
```
kenfdev: Lines above is the main part of the fix. Where it updates the selectedFunction to the new object.

alexellis:
Do you want to update to use $interval?

gateway/assets/index.html
alexellis: Looks like a good change.

gateway/assets/script/bootstrap.js
alexellis: What does `$filter` do? is it the same effect as a for loop and matching on name?

kenfdev: Yes. I wanted to use [find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) but it's a relatively new function so didn't want to break older browsers.

I wondered if I should add a simple `find` function and call that, or to use `$filter` but chose to take advantage of AngularJS's built in function.

kenfdev:
> Do you want to update to use $interval?
To avoid future weird bugs, I'd say yes.  Using the native timeout/interval and updating the objects outside of the AngularJS context may lead to properties not getting updated on the UI.
It has no problems right now because, we're using the `$http` service. Which is aware of the AngularJS context.

Issue #537, opened 2018-02-25
UI may be broken in IE11


alexellis:
I tested the UI using a Microsoft free VM via https://github.com/openfaas/faas/blob/master/contrib/HACK.md

It seemed to work in "Edge" but fail in IE11 on Windows 10 - it could be the VM, or this could be a wider issue with IE11.

Either way I wanted to log it.

## Possible Solution

Work through issues in dev console.

## Steps to Reproduce (for bugs)

1. Deploy OpenFaaS somewhere (not in the VM)
2. Navigate to gateway
3. Page failed to load for me

<img width="1011" alt="screen shot 2018-02-25 at 12 26 36" src="https://user-images.githubusercontent.com/6358735/36641586-cb3ac3ae-1a29-11e8-95fc-f2892562cb95.png">

PR #588, opened 2018-03-19
Fix Download for IE11

iyovcheva:
This fixes Invoke/Download errors on IE11

## Motivation and Context

[#537](https://github.com/openfaas/faas/issues/537)

## How Has This Been Tested?

* IE11
* Safari 11.0.3
* Firefox 58.0.2
* Chrome  64.0.3282.186 
* Microsoft Edge 40.15063.0.0

## Types of changes

## Checklist:

### DIFF
--- a/gateway/assets/script/bootstrap.js
+++ b/gateway/assets/script/bootstrap.js
@@ -320,8 +320,9 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$interval', '$f
 ]);
 
 function uuidv4() {
+    var cryptoInstance = window.crypto || window.msCrypto; // for IE11
     return ([1e7]+-1e3+-4e3+-8e3+-1e11).replace(/[018]/g, function(c) {
-      return (c ^ crypto.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16)
+      return (c ^ cryptoInstance.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16)
     })
   }
   
--- a/gateway/assets/script/bootstrap.js
+++ b/gateway/assets/script/bootstrap.js
@@ -87,40 +87,22 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$interval', '$f
 
             var tryDownload = function(data, filename) {
                 var caught;
-
-                var linkElement = document.createElement('a');
+            
                 try {
                     var blob = new Blob([data], { type: "binary/octet-stream" });
-                    var url = window.URL.createObjectURL(blob);
-
-                    linkElement.setAttribute('href', url);
-                    linkElement.setAttribute("download", filename);
-         
-                    var clickEvent;
-
-                    if((navigator.userAgent.indexOf("MSIE") != -1 ) || (!!document.documentMode == true )) { // for IE 11
-                        clickEvent = document.createEvent("MouseEvent");
-
-                        clickEvent.initMouseEvent("click",  /* eventName */
-                                            true,           /* bubbles */
-                                            false,          /* cancelable */
-                                            window,         /* view */
-                                            0,0,0,0,0,      /* detail, screenX, screenY, clientX, clientY */
-                                            false,          /* ctrlKey */
-                                            false,          /* altKey */
-                                            false,          /* shiftKey */
-                                            false,          /* metaKey */
-                                            0,              /* button */
-                                            null            /* relatedTarget */
-                                        );
-                    } else {
-                        clickEvent = new MouseEvent("click", {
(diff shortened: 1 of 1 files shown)


gateway/assets/script/bootstrap.js
```diff
@@ -320,8 +320,9 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$interval', '$f
 ]);
 
 function uuidv4() {
+    var cryptoInstance = window.msCrypto ? window.msCrypto : crypto;
```
alexellis: This was my input.. @kenfdev does it look OK to you?

gateway/assets/script/bootstrap.js
```diff
@@ -96,11 +96,18 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$interval', '$f
                     linkElement.setAttribute('href', url);
                     linkElement.setAttribute("download", filename);
          
-                    var clickEvent = new MouseEvent("click", {
-                        "view": window,
-                        "bubbles": true,
-                        "cancelable": false
-                    });
+                    var clickEvent;
```
alexellis: @iyovcheva please could you explain the new change here?

What do the magic numbers / flags mean or do when combined? i.e.

`"click",true,false,window,0,0,0,0,0,false,false,false,false,0,null`

gateway/assets/script/bootstrap.js
iyovcheva: This should also work

```
var cryptoInstance = window.msCrypto || crypto;
```

gateway/assets/script/bootstrap.js
iyovcheva: @alexellis 

```
object.initMouseEvent (eventName, bubbles, cancelable, view, detail, screenX, screenY, clientX, clientY, ctrlKey, altKey, shiftKey, metaKey, button, relatedTarget);
```

will add a comment in code

gateway/assets/script/bootstrap.js
kenfdev: Both would work but I'd go with 

```
var cryptoObj = window.crypto || window.msCrypto; // for IE 11
```

as the MDN docs says.

https://developer.mozilla.org/en-US/docs/Web/API/Window/crypto

gateway/assets/script/bootstrap.js
iyovcheva: Addressed. Modification is on another line.

gateway/assets/script/bootstrap.js
alexellis: I tried that approach and it failed to work in IE11 giving an exception, this is how we ended up with the above.

iyovcheva: @alexellis I did a test with the latest changes on all 4 listed browsers&versions. What was the exception?

kenfdev:
@iyovcheva You should use `navigator.msSaveOrOpenBlob` to hack the IE way. The code below should work. (Tested in IE11 and Chrome)

```
var tryDownload = function(data, filename) {
    var caught;

    try {
        var blob = new Blob([data], { type: "binary/octet-stream" });

        if (window.navigator.msSaveBlob) { // // IE hack; see http://msdn.microsoft.com/en-us/library/ie/hh779016.aspx
            window.navigator.msSaveOrOpenBlob(blob, filename);
        }
        else {
            var linkElement = window.document.createElement("a");
            linkElement.href = window.URL.createObjectURL(blob);
            linkElement.download = filename;
            document.body.appendChild(linkElement);
            linkElement.click();
            document.body.removeChild(linkElement);
        }

    } catch (ex) {
        caught = ex;
    }
    return caught;
}
```

FYI
https://stackoverflow.com/a/46233123/6822663

alexellis:
Please could you push up a new gateway image for me to test again on IE11 with your specific fix?

kenfdev:
@alexellis pushed it to `kenfdev/gateway:latest-dev` . This image contains the fix and I PRed it here.
https://github.com/iyovcheva/faas/pull/1

iyovcheva:
Thanks @kenfdev 
I can also test within a day and confirm the fix.

iyovcheva:
Derek remove label: no-dco

alexellis:
Hang on.. why did you ask Derek to remove the label?

iyovcheva:
@kenfdev @alexellis Tested the changes, works well

alexellis:
OK since you've both given sign-off on testing and the changes I'll now merge.

alexellis:
Merged - thank you for the help with this.

kenfdev:
Thank you for testing and reviewing!

PR #610, opened 2018-03-27
Replace tabs with spaces

marcosnils:
Signed-off-by: Marcos Lilljedahl <marcosnils@gmail.com>

## Description
compose file is formatted with spaces except for that line


alexellis:
If you feel like contributing to the project we have lots of areas where help would be appreciated.

Issue #664, opened 2018-04-24
Proposal: Allow deploy_stack* scripts to run sudo docker commands


nishakm:
Allow deploy_stack scripts to invoke sudo for users who have not added their user to the docker group. This makes the deploy scripts more user friendly.

## Expected Behaviour
./deploy_stack.sh should work successfully for users who run docker commands with sudo

## Current Behaviour
./deploy_stack.sh fails with permission denied error

## Proposed Change
Check if a user is in the docker group or if the user is root. If yes then invoke docker commands as usual. If not then prepend sudo.

## Context
I am one of those users that runs docker using sudo. This has caused some delay when running the OpenFaaS workshop.

## Environment
* Docker version 18.03.0-ce, build 0520e24

* Using Docker swarm

* Operating System: Fedora 27

BurtonR:
haha! "one of those users..." :smile: 

This sounds like a reasonable request. Perhaps add a check in the deploy script to check if the current user is in the `docker` group or already executing as sudo, and if not, request sudo access?

@nishakm Would you be interested in implementing this change? We are always welcome to pull requests! Also, if you're interested, join our Slack Community: https://docs.openfaas.com/community/

nishakm:
@BurtonR I can submit a PR. I have already sent a request to add me to the Slack Community.

BurtonR:
Awesome! Be sure to have a look at the [Contribution Guide](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) and don't be afraid to ask questions!

alexellis:
@BurtonR please see my comments on #665.

alexellis:
In the same way that `apt-get install` fails to work with an unprivileged account I think the solution (without making the shell script more complicated)  is to type in `sudo ./deploy_stack.sh` just like `sudo apt-get install`. This is explained in more detail on the PR.

nishakm:
As stated in the proposal, the motivation is to make them user friendly to users who do not add their user to the docker group or do not run as root. The obvious reason for not running docker commands as root or with root-like privileges is security. Docker's documentation specifically warns about this: https://docs.docker.com/install/linux/linux-postinstall/

[...]

PR #665, opened 2018-04-25
Enable deploy_stack scripts to use sudo

nishakm:
Some users do not run docker commands as sudo. This enables deploy_stack.sh and deploy_stack.armhf.sh to check if the user is root or if the user is in the 'docker' group. 

## Description
A new script was added to check if a user has privileges to run docker commands. It checks to see if the user is root. If the user is not root it will check to see if the user is in the 'docker' group. If either one of these conditions are met the script exists with 0. Otherwise it exits with 1.
The deploy_stack.sh and the deploy_stack.armhf.sh then use the check_user.sh script to check if the user has privileges or not and then prepends 'sudo' to the command accordingly

## Motivation and Context
I have raised an issue to propose this change: https://github.com/openfaas/faas/issues/664

## How Has This Been Tested?
I have not tested the deploy_stack.armhf.sh. To test the deploy_stack.sh, I added my user to the docker group and ran the script.

## Types of changes

## Checklist:

### DIFF
--- a/check_user.sh
+++ b/check_user.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+# Check if a user is root or is in the docker group
+# The exit code is 0 if the user has permissions
+# to run docker commands
+# The exit code 1 if the user has to run docker
+# commands using sudo
+
+retval=1
+if [ $(id -u) == 0 ]; then
+  retval=0
+else
+  for group in $(groups)
+  do
+    if [ $group == 'docker' ]; then
+      retval=0
+      break
+    fi
+  done
+fi
+exit $retval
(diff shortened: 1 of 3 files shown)


alexellis:
Hi Nisha,

Thanks for making this suggestion - I don't think it's one that's come up before. We do encourage new contributors to raise a proposal issue for discussion *before going ahead and working on a solution*.

I am wondering what the use-case is for not adding one's user-account to the Docker group? This is the [de-facto](https://docs.docker.com/install/linux/linux-postinstall/#configure-docker-to-start-on-boot)/recommended way to use and operate Docker on Mac/Linux and users on Windows need to run an elevated cmd/ps Console to use the CLI.

I wonder if for your use-case you could just type in `sudo ./deploy_stack.sh` in a similar way to how one would run `apt-get install`. What do you think - could that work?

Alex

nishakm:
I raised #664 before working on a solution (darn timezones!). I'll comment there from here on out.

alexellis:
Derek close: marked for #revisit. Thank you for the PR and for using the project 👍

Issue #238, opened 2017-09-29
Research item: Scale to zero and back again


alexellis:
This is a research item or PoC - not necessarily going to be merged into the project. The aim is to learn about challenges and trade-offs about scaling to zero and back again.

## Expected Behaviour

Given a service has been scaled to zero replicas, upon synchronous or asynchronous invocation we should increase the replica count to 1/1 and then wait until the function is ready and invoke it.

## Current Behaviour

Scaling to zero is the same as "not found"

## Possible Solution

Ideas:
* Implement spin-down service by watching Prometheus metrics
* Add attribute/meta to functions which can spin down or up
* Implement smart-proxy to spin up a function from 0/1 to 1/1 replicas and poll until ready

## Context

This should be implemented at the FaaS repo level so that it can be utilised across all providers.

sorenbs:
What is the best case delay when scaling up from zero?

~ 10 ms
~ 100 ms
~ 1 s
~ 10 s

alexellis:
That's a great question, but it depends on so much - including the back-end you're targeting and the hardware you've deployed to. Are you adding to the discussion or keen to try implementing this?

wonderxboy:
Anyone working on this？

alexellis:
Yes I have a prototype but Docker Swarm is misrepresenting readiness so going back up to 1/1 was less reliable

wonderxboy:
@alexellis I see your code proxy.go only spins up the service instance when there is 0 instance of such a service. 
1. How could a service get into a 0-instance at the first place? At deploy time with a instance number(faas-cli does not seem to have such an option)? 
2. Does it scale back to 0 automatically when there is no activity for a given time span?

alexellis:
You would use the logs to decide when the function was inactive and then scale down.

alexellis:
Logs or metrics.

wonderxboy:
Would you work on that? Or should I have a look? I would expect this to be a full-fledged feature in that the system automatically take care of the whole process. This is a very useful feature form a cloud provider's perspective, the functions can be served with less resource consumptions

BurtonR:
Bumping this as it's a common question when I do demonstrations of OpenFaaS to the community. 

@wonderxboy or @alexellis has any progress been made on this? I'd love to help if I can 😄

alexellis:
[...]

PR #685, opened 2018-05-14
Add feature: scale from zero to 1 replicas

alexellis:
## Description

This change allows functions to be "idled" or scaled to zero
replicas and then be invoked later on. There is a penalty to
scaling up - the API gateway proxy will block until the function
is ready.

A cache is included to off-set the calls to upstream API to check
on readiness along with unit tests.

## Motivation and Context

Addresses part of #238 

## How Has This Been Tested?

Testing via scaling to zero replicas and then invoking function.
On Swarm I observed 3 seconds on an Intel Nuc i5 for scaling back
from zero replicas.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/handlers/scaling.go
+++ b/gateway/handlers/scaling.go
@@ -1,0 +1,147 @@
+// Copyright (c) OpenFaaS Project. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package handlers
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"time"
+
+	"github.com/openfaas/faas/gateway/requests"
+)
+
+// ScalingConfig for scaling behaviours
+type ScalingConfig struct {
+	MaxPollCount         uint
+	FunctionPollInterval time.Duration
+	CacheExpiry          time.Duration
+}
+
+// MakeScalingHandler creates handler which can scale a function from
+// zero to 1 replica(s).
+func MakeScalingHandler(next http.HandlerFunc, upstream http.HandlerFunc, config ScalingConfig) http.HandlerFunc {
+	cache := FunctionCache{
+		Cache:  make(map[string]*FunctionMeta),
+		Expiry: config.CacheExpiry,
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+
+		functionName := getServiceName(r.URL.String())
+
+		if replicas, hit := cache.Get(functionName); hit && replicas > 0 {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		replicas, code, err := getReplicas(functionName, upstream)
+		cache.Set(functionName, replicas)
+
+		if err != nil {
+			var errStr string
+			if code == http.StatusNotFound {
+				errStr = fmt.Sprintf("unable to find function: %s", functionName)
+
+			} else {
+				errStr = fmt.Sprintf("error finding function %s: %s", functionName, err.Error())
+			}
+
+			log.Printf(errStr)
+			w.WriteHeader(code)
+			w.Write([]byte(errStr))
+			return
+		}
+
+		if replicas == 0 {
+			minReplicas := uint64(1)
+
+			err := scaleFunction(functionName, minReplicas, upstream)
+			if err != nil {
+				errStr := fmt.Errorf("unable to scale function [%s], err: %s", functionName, err)
+				log.Printf(errStr.Error())
+
+				w.WriteHeader(http.StatusInternalServerError)
+				w.Write([]byte(errStr.Error()))
+				return
+			}
+
+			for i := 0; i < int(config.MaxPollCount); i++ {
+				replicas, _, err := getReplicas(functionName, upstream)
+				cache.Set(functionName, replicas)
+
+				if err != nil {
+					errStr := fmt.Sprintf("error: %s", err.Error())
+					log.Printf(errStr)
+
+					w.WriteHeader(http.StatusInternalServerError)
+					w.Write([]byte(errStr))
+					return
+				}
+
+				if replicas > 0 {
+					break
+				}
+
+				time.Sleep(config.FunctionPollInterval)
+			}
+		}
+
+		next.ServeHTTP(w, r)
+	}
+}
+
+func getReplicas(functionName string, upstream http.HandlerFunc) (uint64, int, error) {
+
+	replicasQuery, _ := http.NewRequest(http.MethodGet, fmt.Sprintf("/system/function/%s", functionName), nil)
+	rr := httptest.NewRecorder()
+
+	upstream.ServeHTTP(rr, replicasQuery)
+	if rr.Code != 200 {
+		log.Printf("error, query replicas status: %d", rr.Code)
+
+		var errBody string
+		if rr.Body != nil {
+			errBody = string(rr.Body.String())
+		}
+
+		return 0, rr.Code, fmt.Errorf("unable to query function: %s", string(errBody))
+	}
+
+	replicaBytes, _ := ioutil.ReadAll(rr.Body)
+	replicaResult := requests.Function{}
+	json.Unmarshal(replicaBytes, &replicaResult)
+
+	return replicaResult.AvailableReplicas, rr.Code, nil
+}
+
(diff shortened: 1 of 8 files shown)


alexellis:
Video demo with Swarm: https://www.youtube.com/watch?v=-6DPUFfn1Ss&index=16&t=1s&list=PLlIapFDp305BebghsydcwLYO_zyI1kBVc

alexellis:
@vicary did you want to try this out?

The scale to zero is done via a controller which observes Prometheus metrics for inactivity.

https://github.com/openfaas-incubator/faas-idler

viveksyngh:
@alexellis does it scales down to zero after some period of inactivity?

stefanprodan:
@viveksyngh check https://github.com/openfaas-incubator/faas-idler code

stefanprodan:
@alexellis I would make the "scale to 0" setting per function using `com.openfaas.scale.min: "0"` label.

stefanprodan:
I would also move the logic from faas-idler to the gateway code since the gateway already owns the auto-scaling logic.

alexellis:
@stefanprodan moving the idler into the gateway would require something like an additional go-proc to the http server, do you think that makes sense? We'd also get N replicas of the idler per N replicas of the gateway.

ericstoekl:
I tested this out on swarm and K8s, and it works on both.

On swarm, I used this command to scale a deployed function to 0, then invoked it:

```
docker service scale markdown=0
```

On K8s, I used this command:

```
 kubectl -n openfaas-fn scale --replicas=0 deploy/base64
```

In both cases, after a ~5 second delay, the function returned the expected result. I checked the Swarm service list/kubernetes namespace, and the function had come back to `1/1` replicas.

alexellis:
Thanks for the data @ericstoekl 

I noted 3 seconds on Swarm and 10-15 on Kubernetes with faas-netes.

Please could you look at stress testing via `hey` and also try the faas-idler in combination with that?

ivanayov:
Scale up works as expected. I did the same test as @ericstoekl 

IMO faas-idler is better as a separate service that can be enabled/disabled.
If it's part of the gateway, this can lead to worse performance (especially on production or where the number of functions deployed is big enough to cause a lot of consumption)

Enable/Disable per function is also a good to have.

alexellis:
Thanks for your input @ivanayov 

I think opt-in/out per function is smart, but I'm not sure about the specific idle time per function because that may involve having to make several queries to prometheus for each function - right now with one idle duration we can make a single request to prometheus and then run through the results.

> I think faas-idler should better be a separate service that can be enabled/disabled.

Right now the gateway API doesn't expose set-scale, but the internal provider API does, so we either update the GW API to expose that or move the code into the gateway.

I have mixed feelings about it running in the gateway process because it would need to be a separate go proc with its own logs and timer.

> If it's part of the gateway, this can lead to worse performance (especially on production or where the number of functions deployed is big enough to cause a lot of consumption)

I'm not disagreeing, but this is just a hypothesis - can it be substantiated with data? What is it about the current design that makes you think that?

alexellis:
Related issue for Operator - https://github.com/openfaas-incubator/openfaas-operator/issues/30

stefanprodan:
Having the scaling feature split in 2 projects doesn’t make sense to me. Maybe a better option would be to move the auto-scaling logic along with the scale to 0 to a dedicated service, something like “openfaas-scaler”.  Both auto-scaling and scale to 0 should be opt-in via the function labels so that a user can opt for K8s HPA for one function and Prometheus req/sec for another.

stefanprodan:
And since we need to query the data from Prometheus on a cron, we can drop Alertmanager and make the auto scaling smarter :)

ericstoekl:
In kubernetes, when hitting the function with the following `hey` request, it seems that the function fails to scale from 0 to 1 replicas:

```
$ hey -n 1000 -c 200 -m POST -d 'stuff' $OPENFAAS_URL/function/base64

Summary:
  Total:	100.0534 secs
  Slowest:	19.9951 secs
  Fastest:	10.1903 secs
  Average:	15.2583 secs
  Requests/sec:	9.9947
  
  Total data:	2736 bytes
  Size/request:	57 bytes

Response time histogram:
...
Status code distribution:
  [502]	48 responses

Error distribution:
  [952]	Post http://192.168.99.100:31112/function/base64: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

In Docker Swarm, it seems that the `hey` command sees 500 responses until the function scales back up to 1, then gets 200 responses after that.

```
$ hey -n 10000 -c 200 -m POST -d 'stuff' http://localhost:8080/function/base64

Summary:
  Total:	23.9536 secs
  Slowest:	16.7546 secs
  Fastest:	0.0014 secs
  Average:	0.4639 secs
  Requests/sec:	417.4730
  
  Total data:	306916 bytes
  Size/request:	30 bytes
...
Status code distribution:
  [200]	6444 responses
  [500]	3556 responses

```

What I have observed it that in kubernetes the function will eventually scale up from 0 to 1 replicas, but it will return this error when called directly:

```
curl -d 'stuff' $OPENFAAS_URL/function/base64
error finding function base64: unable to query function: ~
```

berndtj:
I think the scale-up time is hindered by the k8s readiness probe.  We could likely improve that significantly, if we simply didn't depend on it.  Rather have the gateway probe the function /healthz endpoint directly to determine the state of the function.

alexellis:
Eric look at the data:

> Slowest:	19.9951 secs

This is a timeout. The timeout values have to be increased to accommodate the difference introduced by scheduling the Pod and waiting for the health checks to pass.

Try increasing the timeouts to 60s.

gateway/handlers/scaling.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+
+			log.Printf(errStr)
+			w.WriteHeader(code)
+			w.Write([]byte(errStr))
+			return
+		}
+
+		if replicas == 0 {
+			minReplicas := uint64(1)
```
alexellis: `1` should pull min replicas via label -> `com.openfaas.scale.min` or similar.

gateway/handlers/scaling.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	scaleBytesOut, _ := json.Marshal(scaleReq)
+	scaleBytesOutBody := bytes.NewBuffer(scaleBytesOut)
+	setReplicasReq, _ := http.NewRequest(http.MethodPost, fmt.Sprintf("/system/scale-function/%s", functionName), scaleBytesOutBody)
+
+	rr := httptest.NewRecorder()
+	upstream.ServeHTTP(rr, setReplicasReq)
+
+	if rr.Code != 200 {
```
alexellis: Needs to check `200` or `202`.

gateway/server.go
```diff
@@ -113,10 +113,18 @@ func main() {
...
 	r := mux.NewRouter()
+	// max wait time to start a function = maxPollCount * functionPollInterval
 
+	scalingConfig := handlers.ScalingConfig{
+		MaxPollCount:         uint(1000),
+		FunctionPollInterval: time.Millisecond * 10,
+		CacheExpiry:          time.Second * 5, // freshness of replica values before going stale
+	}
+
+	scalingProxy := handlers.MakeScalingHandler(faasHandlers.Proxy, queryFunction, scalingConfig)
```
alexellis: I think this should be feature flagged before releasing - i.e. env-var: `scale_from_zero: true` - we might also need to load the "ScalingConfig" via env-vars at some point in the future. Let's see how useful the defaults are first.

ivanayov:
@alexellis I'm working on addressing the comments

alexellis:
( Context is helpful for the large thread - Ivana is specifically looking at my review comments above on my PR. )

berndtj:
That's certainly the easy solution.  I'm also currently playing with creating a different readiness check (it's just an image after all) which would retry on connection errors before returning (up to some timeout).  That would give the most control over function probes.  However, I can certainly issue a PR with the changes to the polling as a quick fix.  I don't think there are any real downsides to polling more frequently.

alexellis:
For testing please use `openfaas/gateway:scale-17-07-2018` Docker image and set env-var `scale_from_zero` to true.

jahentao:
It seems not works on Nomad with faas-nomad.

alexellis:
Feel free to raise an issue with Hashicorp. We cover Kubernetes, Swarm and faasd (containerd) which are all working as expected.

alexellis:
https://github.com/hashicorp/faas-nomad

Issue #687, opened 2018-05-18
Proposal: Move Basic Auth into gateway code


alexellis:
We currently secure the OpenFaaS administrative API route `/system/` by fronting it with Basic Auth through a HTTP Reverse Proxy.

If we implement basic auth within the the `/system/` route we can decouple this implementation detail.

The implementation may require the user to create a secret using `kubectl` / `docker swarm` before deploying.

We should also look at alternative options.

Do not forget faas-netesd / openfaas-operator / faas-swarm.

alexellis:
Test

alexellis:
Some interesting ideas on different auth strategies: https://facundoolano.wordpress.com/2013/12/23/my-take-on-restful-authentication/

cpitkin:
@alexellis Which auth strategy in that list did you have in mind? Starting out with Basic Auth then adding more later might not be a bad idea.

alexellis:
Would like input from the community on this. We've seen basic auth offer usefulness and simplicity, but is there more we could do?

PR #691, opened 2018-05-21
Enable basic auth on gateway

alexellis:

## Description

Fixes https://github.com/openfaas/faas/issues/687 allowing the
gateway to handle the responsibility of basic auth for when it is
in use.

To enable set basic_auth env-var to true and then mount two
secrets or plaintext files under /var/secrets/

basic_auth_user, basic_auth_password

## Motivation and Context

#687

## How Has This Been Tested?

Tested with faas-cli list/deploy and with Safari browser.

## Types of changes

## Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,8 @@ services:
             faas_nats_port: 4222
             direct_functions: "true"    # Functions are invoked directly over the overlay network
             direct_functions_suffix: ""
+            basic_auth: "false"
+            secret_mount_path: "/run/secrets/"
         deploy:
             resources:
                 # limits:   # Enable if you want to limit memory usage
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -5,8 +5,10 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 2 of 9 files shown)


gateway/server.go
```diff
@@ -33,6 +35,27 @@ func main() {
 
 	log.Printf("Binding to external function provider: %s", config.FunctionsProviderURL)
 
+	var credentials *handlers.BasicAuthCredentials
+
+	if config.UseBasicAuth {
+		userPath := "/var/secrets/basic_auth_user"
```
johnmccabe: You need to check the `/run/secrets/*` files when running on Swarm or the gateway will fail to initialise.

alexellis: In which case should I make the secret path a config item too?

johnmccabe: Could just check both, its only going to occur once on startup so shouldn't be an issue? Saves having to expose more config to the user.

alexellis: @stefanprodan is there such a thing as a standard location for secrets on Kubernetes?

stefanprodan: There is no  standard location, usually ok Kubernetes you would mount secrets as ENV vars not files on disk.

alexellis:
@johnmccabe I think I've addressed your concern now. It is not on by default since we need to figure out how to have the user create the secrets first.

docker-compose.yml
```diff
@@ -16,6 +16,8 @@ services:
             faas_nats_port: 4222
             direct_functions: "true"    # Functions are invoked directly over the overlay network
             direct_functions_suffix: ""
+            basic_auth: "false"
```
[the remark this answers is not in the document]
alexellis: I think to use it you would need to do that, but we'd not be able to do a stack deploy unless these secrets existed.

alexellis:
Thank you guys for reviewing and for your feedback.

Issue #692, opened 2018-05-22
Proposal: Change secrets to mount in a specialized subfolder


LucasRoesler:
To address various issues that we have recently seen occur with secrets (mostly in Kubernetes) and to maintain cross-compatibility between Swarm and Kubernetes, we should mount all secrets to  `/var/openfaas/secrets`.

## Expected Behaviour

When a user indicates that a secret is required, we should specify that the orchestration layer mounts that secret inside `/var/openfaas/secrets`.  For example, in Swarm, if the user has a secret `postgresuser`, then inside the function container, they secret file would be found at `/var/openfaas/secrets/postgresuser`.

Internally, the faas provider would be setting the mount point via their respective APIs:

* for Swarm, we would use the `source=%s,target=%s` notation as described here https://docs.docker.com/engine/swarm/secrets/#intermediate-example-use-secrets-with-a-nginx-service
* for Kubernetes this would require setting the `mountPath` appropriately, see https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod

## Current Behaviour

This behavior mimics the current behavior of mounting the secrets at `/run/secrets` but will avoid conflicts with how other systems expect that folder to behave.  

## Context

This provides a solution to bugs like openfaas/faas-netes#188 in which the Kubernetes cluster fails to launch the function container because there is a permissions issue with the `/run/secrets` folder.

One of the goals in this proposal is to continue to preserve a consistent experience between Kubernetes and Swarm for function authors.  Currently, and continuing with this proposal, secrets will also be available in the same location and assuming that they do not use multiple files per Kubernetes secret, then the function can access the secrets in exactly the same way in both Kubernetes and Swarm.

LucasRoesler:
Some comments on the actual implementation, I am pretty sure that for both swarm and k8s, this is a one line change.  

In swam, we would just need to update https://github.com/openfaas/faas-swarm/blob/master/handlers/deploy.go#L345 to use a formatted string with the `source=%s,target=/var/openfaas/%s` instead of the secret name, like it currently does.  Essentially we would be piggybacking on the swarm cli parsing, which we already do anyway. 

In k8s, we would simply update the `mountPath` here https://github.com/openfaas/faas-netes/blob/master/handlers/secrets.go#L98

LucasRoesler:
[...]

PR #702, opened 2018-05-26
Update secret docs and sample-functions

LucasRoesler:
## Description
- Update the documentation about secret management to note the changed
file location
- Remove the documentation on secret rotation, because this will not
currently work
- Update apikey-secret and ApiKeyProtected-Secrets to read secret values
from both the old and the new locations

## Motivation and Context

Recent updates to faas-swarm and faas-netes changed the mount location
of secrets to implement #692.  These changes update the docs to reflect this change.

## Types of changes

## Checklist:

### DIFF
--- a/guide/secure_secret_management.md
+++ b/guide/secure_secret_management.md
@@ -4,7 +4,8 @@ OpenFaaS deploys functions as Docker Swarm Services, as result there are several
 
 ## Using Environment Variables
 
-First, and least secure, is the ability to set environment variables at deploy time. For example, you might want to set the `NODE_ENV` or `DEBUG` variable.  Setting the `NODE_ENV` in the stack file `samples.yml`
+First, and least secure, is the ability to set environment variables at deploy time. For example, you might want to set the `NODE_ENV` or `DEBUG` variable. Setting the `NODE_ENV` in the stack file `samples.yml`
+
 ```yaml
 provider:
   name: faas
@@ -20,6 +21,7 @@ functions:
 ```
 
 You can then deploy and invoke the function via the `faas-cli` using
+
 ```sh
 $ faas-cli deploy -f ./samples.yml
 $ faas-cli invoke nodehelloenv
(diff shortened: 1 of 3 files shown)


guide/secure_secret_management.md
```diff
@@ -87,11 +88,11 @@ functions:
     image: functions/api-key-protected:latest
```
alexellis: Do we need a PR to this function and for me to release it to the hub?

guide/secure_secret_management.md
```diff
@@ -87,11 +88,11 @@ functions:
...
 ```
 
-Note that unlike the `envVars` in the first example, we do not provide the secret value, just a list of names: `"secrets": ["secret_api_key"]`. The secret value has already been securely stored in the Docker swarm.  One really great result of this type of configuration is that you can simplify your function code by always referencing the same secret name, no matter the environment, the only change is how the environments are configured.
+Note that unlike the `envVars` in the first example, we do not provide the secret value, just a list of names: `"secrets": ["secret_api_key"]`. The secret value has already been securely stored in the Docker swarm. One really great result of this type of configuration is that you can simplify your function code by always referencing the same secret name, no matter the environment, the only change is how the environments are configured.
```
alexellis: What changed on these lines? I can't see from the diff.

guide/secure_secret_management.md
```diff
@@ -87,11 +88,11 @@ functions:
...
 
-## Advanced Swarm Secrets
+<!-- ## Advanced Swarm Secrets
```
alexellis: Is this commented out?

alexellis:
We'll also need to update https://docs.openfaas.com via https://github.com/openfaas/docs/

guide/secure_secret_management.md
rorpage: Looks like the removal of an extra space

johnmccabe: screen shot 2018-05-31 at 23 49 56

sample-functions/ApiKeyProtected-Secrets/handler.go
```diff
@@ -8,10 +8,21 @@ import (
 	"strings"
 )
 
+func getAPISecret(secretName string) (secretBytes []byte, err error) {
```
johnmccabe: I would be inclined to swallow the error logging into the function rather than have the user deal with it when getting the secret:
```
// getSecret returns the secret found at the first valid secret mountpoint
func getSecret(name string) []byte {
	mounts := []string{"/var/openfaas/secrets/", "/run/secrets/"}
	var b []byte
	var err error
	for _, m := range mounts {
		if b, err = ioutil.ReadFile(m + name); err == nil {
			return b
		}
	}
	log.Fatal(err)
	return nil
}
```
```
secretBytes := getSecret("secret_api_key")
```

guide/secure_secret_management.md
LucasRoesler: yes, this isn't supported anymore.  Before we merged the updated behavior, i had brought this up, but the new behavior was merged first

LucasRoesler:
I created a PR in the docs project,. The style of the refernce doc is a bit different, so it is not a simple copy of this pr.

alexellis:
Remaining item is to update docker-compose.yml to reflect the newer faas-swarm version which contains this change. K8s is already up to date then we can merge this and the docs PR.

alexellis:
@LucasRoesler

alexellis:
Yes I did it before heading into the plane. I'll merge this PR now too. Is the workshop repo up to date too?

Issue #697, opened 2018-05-24
Proposal: set replicas to zero in Prometheus when deleting a function


CuZn13:
# Title

When deleting a function, notify prometheus to set replicasCounter to 0

Delete a function, but still show one in prometheus. My approach is to add a deleteProxyHandler  add a method to metrics.go. Set replicasCounter to 0 when deleting a function

`func MakeDeleteFunctionProxyHandler(proxy *types.HTTPClientReverseProxy, metrics metrics.MetricOptions, baseURLResolver BaseURLResolver) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		baseURL := baseURLResolver.Resolve(r)
		body, _ := ioutil.ReadAll(r.Body)
		req:=requests.DeleteFunctionRequest{}
		json.Unmarshal(body, &req)
		delete(proxy.Client,baseURL,req.FunctionName,metrics)
	}
}

func delete(proxyClient *http.Client,baseURL string,functionName string,metrics metrics.MetricOptions){
	str :=make(map[string]string)
	str["functionName"]=functionName
	reqBytes, _ := json.Marshal(&str)
	reader := bytes.NewReader(reqBytes)
	request, _ := http.NewRequest(http.MethodDelete, baseURL+"/system/functions",reader)
	request.Header.Set("Content-Type", "application/json")
	_,err:=proxyClient.Do(request)
	if err==nil{
		log.Printf("delete functon [%s] is ok",functionName)
		trackFunctionStop(metrics,functionName)
	}
}

func trackFunctionStop(metrics metrics.MetricOptions, name string) {
	metrics.ServiceReplicasCounter.
		WithLabelValues(name).
		Set(float64(0))
}`

alexellis:
Hi I think this sounds like a good approach. Thanks for using OpenFaaS and for logging the issue. 

Let's get another opinion and then look at moving forward with a solution and fix. 

Alex

kenfdev:
Thank you for pointing this out @CuZn13 . This looks nice.  In addition, imo we should continue using the logic of the `MakeForwardingProxyHandler` so that we don't break the original functionality and not to duplicate code.

Something like the following:

```
func MakeDeleteFunctionProxyHandler(proxy *types.HTTPClientReverseProxy, notifiers []HTTPNotifier, baseURLResolver BaseURLResolver, metricsOptions metrics.MetricOptions) http.HandlerFunc {
[...]

PR #705, opened 2018-05-28
Set replicas to zero in Prometheus when deleting a function

kenfdev:
Signed-off-by: Ken Fukuyama <kenfdev@gmail.com>

## Description

Currently, when deleting a function the replicas count in Prometheus doesn't get set to 0. This PR sets the replica count to 0 after forwarding the DELETE request to the upstream provider.

## Motivation and Context

#697 

## How Has This Been Tested?

This has been tested on Docker Swarm.

```
* Create Function
* Confirm replica count is `1` in Prometheus
* Delete Function
* Confirm replica count is `0` in Prometheus
```

```
* Create Function
* Scale the function to 5 replicas
* Confirm replica count is `5` in Prometheus
* Delete Function
* Confirm replica count is `0` in Prometheus
```

Sample Prometheus Capture:
image

The container image is pushed at `kenfdev/gateway:latest-dev`.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/handlers/deletefunction_proxy.go
+++ b/gateway/handlers/deletefunction_proxy.go
@@ -1,0 +1,67 @@
+package handlers
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/openfaas/faas/gateway/metrics"
+	"github.com/openfaas/faas/gateway/requests"
+	"github.com/openfaas/faas/gateway/types"
+)
+
+// MakeDeleteFunctionProxyHandler creates a handler which forwards HTTP requests and sets replicasCount to 0
+func MakeDeleteFunctionProxyHandler(proxy *types.HTTPClientReverseProxy, notifiers []HTTPNotifier, baseURLResolver BaseURLResolver, metricsOptions metrics.MetricOptions) http.HandlerFunc {
+
+	return func(w http.ResponseWriter, r *http.Request) {
+
+		functionName, parseErr := getFunctionNameFromRequest(r)
+		if parseErr != nil {
+			log.Printf("Error parsing request to remove service.")
+		}
+
+		baseURL := baseURLResolver.Resolve(r)
+
+		requestURL := r.URL.Path
+
+		start := time.Now()
+
+		statusCode, err := forwardRequest(w, r, proxy.Client, baseURL, requestURL, proxy.Timeout)
+
+		seconds := time.Since(start)
+
+		if err != nil {
+			log.Printf("Error with upstream request to: %s, %s\n", requestURL, err.Error())
+		} else {
+			log.Printf("Setting replicas count to 0 for %s\n", functionName)
+			trackFunctionStop(metricsOptions, functionName)
+		}
+		for _, notifier := range notifiers {
+			notifier.Notify(r.Method, requestURL, statusCode, seconds)
+		}
+	}
+}
+
+func getFunctionNameFromRequest(r *http.Request) (string, error) {
+	req := requests.DeleteFunctionRequest{}
+	reqData, _ := ioutil.ReadAll(r.Body)
+
+	unmarshalErr := json.Unmarshal(reqData, &req)
+	if unmarshalErr != nil {
+		log.Printf("Error unmarshaling the request: %s\n", unmarshalErr.Error())
+		return "", unmarshalErr
+	} else if len(req.FunctionName) == 0 {
+		log.Printf("The function name in the request was empty\n")
+		return "", fmt.Errorf("Function name empty")
+	}
+
+	// Restore the io.ReadCloser to its original state so the body is readable
+	// when forwarding the request
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(reqData))
+
+	return req.FunctionName, nil
+}
--- a/gateway/handlers/delete_proxy_test.go
+++ b/gateway/handlers/delete_proxy_test.go
@@ -1,0 +1,170 @@
+package handlers
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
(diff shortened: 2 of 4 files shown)


gateway/handlers/deletefunction_proxy.go
```diff
@@ -0,0 +1,67 @@
...
+			log.Printf("Error parsing request to remove service.")
+		}
+
+		baseURL := baseURLResolver.Resolve(r)
+
+		requestURL := r.URL.Path
+
+		start := time.Now()
+
+		statusCode, err := forwardRequest(w, r, proxy.Client, baseURL, requestURL, proxy.Timeout)
```
kenfdev: I've decided to reuse the previous functionality for DELETE as much as possible. So am reusing the `forwardRequest`

gateway/handlers/deletefunction_proxy.go
```diff
@@ -0,0 +1,67 @@
...
+		log.Printf("Error unmarshaling the request: %s\n", unmarshalErr.Error())
+		return "", unmarshalErr
+	} else if len(req.FunctionName) == 0 {
+		log.Printf("The function name in the request was empty\n")
+		return "", fmt.Errorf("Function name empty")
+	}
+
+	// Restore the io.ReadCloser to its original state so the body is readable
+	// when forwarding the request
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(reqData))
```
kenfdev: A bit hackish but in order to reuse the `forwardRequest` and *also* read the function name from the request Body, the `r.Body` needs to be reassigned (it needs to be read 2 times).

kenfdev:
@alexellis @johnmccabe @rgee0 @stefanprodan Would greatly appreciate your feedbacks when you have time. Thanks in advance.

rorpage:
I think this looks great! I have a couple questions out of curiosity but :+1:

ericstoekl:
I tested this out and it works -- I see the replica count in Prometheus dropping to 0 after the function is deleted.

Also, code coverage is close to 100% on the newly added code. Good work!

LGTM

kenfdev:
@ericstoekl  @rorpage Thanks for looking into this.

@rorpage 
Also, would love to hear what you were curious about 😉

rorpage:
@kenfdev I had added some comments above ^

gateway/handlers/deletefunction_proxy.go
alexellis: Not sure of this. I need to look into the implementation to understand how you got here.

alexellis:
I think it would make the code more maintainable if we used a middleware wrapper for this. Look at how callId is injected. Could this work too ?

kenfdev:
> more maintainable if we used a middleware wrapper for this

@alexellis I agree with this and was trying to take this approach at first.  But since the `forwardRequest` consumes the `r.Body`, the middleware cannot read the functionName anymore (because the stream is empty). We can tweek it by reading the function name and then call `next(w, r)`, and finally setting prometheus metrics like the pseudo code below:

```
// pseudo code
func MakeDeleteMiddleware(next http.HandlerFunc) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		// read function name and re-set the r.Body
		fnName := getFunctionNameFromRequest(r)

                 // forward the request
		next(w, r)

                 // set prometheus metrics
		setPrometheusReplicasToZero(fnName)
	}
}
```

WDYT?

kenfdev:
One more thing about the middleware approach is that, currently there's no way to know if the forward request succeeded or not. So the replicas count will always be set to zero even if the delete request fails. (That's why I'm using `forwardRequest` in this PR to be able to look at the `err`)

```
// pseudo code
func MakeDeleteMiddleware(next http.HandlerFunc) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		// read function name and re-set the r.Body
		fnName := getFunctionNameFromRequest(r)

                 // forward the request
		next(w, r)  // cannot determine whether the request succeeded or not

                 // therefore, "set prometheus metrics" will always fire
		setPrometheusReplicasToZero(fnName) 
	}
}
```

BUT, at the end, the `externalwatcher` will re-set the replicas count so eventually the replicas count would be set to the proper count. I think it's a design decision.

@LucasRoesler Do you have any opinions about this? Thanks!

LucasRoesler:
Getting the request status code in a middleware, while not possible in stdlib, it is a common pattern implemented (over and over again unfortunately) in many logging middlewares.  The negroni middleware is nice for this, because it actually exports the needed `ResponseWriter` type.  See https://github.com/urfave/negroni/blob/master/response_writer.go#L13 for the definition and https://github.com/urfave/negroni/blob/master/logger.go#L63 for the usage.  

I also like the idea of a middleware and I think using this would resolve the status code issue.

alexellis:
I'm updating myself on this PR. What if we also had the function name inside the route? Should we imperatively zero metrics when the series is no longer in scope or just expire the data? I.e. a reading is only valid if updated in last N seconds?

I'd like to know what a Prometheus expert would think to this problem.

kenfdev:
Thanks for the suggestions Lucas. The negroni middleware sounds pretty nice. I like how we could extract the status code from the wrapped response writer.

For this specific case, as Alex suggests, adding the function name in the route would be pretty natural as a REST endpoint. I think DELETE requests usually include the ID of the resource in the route. This would be a breaking change though.

> Should we imperatively zero metrics when the series is no longer in scope or just expire the data?

I'm not sure about the practices for Prometheus so would like to know the Prometheus way myself, too.

alexellis:
Having consulted with a Prometheus co-founder I am not sure that our Prometheus implementation is suited to this application. It would be better for us to implement the collector-style interface. Please see here for an example: https://github.com/alexellis/build_executor_exporter/blob/master/app.go

kenfdev:
Thanks for the advices Alex, I'm going to close this PR. The Collector and Describe interface are the right way to implement this. I'll create a different PR for this.

Issue #714, opened 2018-06-06
Functions are accessible if we add slash (/) at the end of url


tarunmangukiya:
Let's say our function name is `download`, and we access that function using GET `http://localhost:8080/function/download` then it gives proper response. But if we access it using `http://localhost:8080/function/download/` then it throws 502 server error.

## Expected Behaviour
I think both of the urls are technically same and both of them should work.

## Current Behaviour
It throws 502 server error. I've checked gateway logs and found that gateway upstreams it to container as `http://download/:8080/function/download`, there is trailing slash.

## Possible Solution

Both urls should work.

## Steps to Reproduce (for bugs)

1. Create any function which provides response in GET request, example, `download`
2. Access it from `http://localhost:8080/function/download`
3. Access it from `http://localhost:8080/function/download/`
4. Check the outputs

## Context

I was trying to keep my server env file neat. There were slash at the end in some urls, as we're using them for internal proxying and http requests. So, I created all the urls with slash at the end.

affix:
I don't feel this is a bug, more of a proposal.

However I don't see a real benefit to supporting a trailing slash in the URL as this is something that can be handled by your load balancing software (if using any).

A trailing slash would break get params and theres no real benefit to supporting both URL schemed.

making it "neat" doesn't seem like a valid reason for this change.

alexellis:
Can we validate this with the normal classic watchdog? And the latest gateway.

Which docker image are you using?

alexellis:
I reproduced the problem on Kubernetes with the nodeinfo function. It used to work with a trailing slash, so it should still do that now. I think it may be a regression around the direct functions access code. I'd like help looking into this.

LucasRoesler:
I got it

affix:
Tested this with the latest gateway on k8s using figlet

```
➜  ~ curl -d "Hello World" https://<hidden>/function/figlet
 _   _      _ _        __        __         _     _
| | | | ___| | | ___   \ \      / /__  _ __| | __| |
| |_| |/ _ \ | |/ _ \   \ \ /\ / / _ \| '__| |/ _` |
|  _  |  __/ | | (_) |   \ V  V / (_) | |  | | (_| |
|_| |_|\___|_|_|\___/     \_/\_/ \___/|_|  |_|\__,_|

➜  ~ curl -d "Hello World" https://<hidden>/function/figlet/
 _   _      _ _        __        __         _     _
| | | | ___| | | ___   \ \      / /__  _ __| | __| |
[...]

PR #717, opened 2018-06-06
[WIP] Trim tailing slash from service name before proxy

LucasRoesler:
## Description

When determining the service name of the function, remove any trailing
slashes, the slashes are not allowed in service names for either Swarm
or K8S, so this can only be a left over from the url path

## Motivation and Context

- This was preventing service resolution, and hence failed functions,
    when the function was called with a trailing slash. See #714

## How Has This Been Tested?

I have added a unit test, I have not tested a deploy yet because it is getting late here, but I wanted to get something posted incase someone else was feeling like testing.

I did run my test prior to update the method, and they were failing with the trailing slash.  They now pass. 

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/gateway/handlers/forwarding_proxy.go
+++ b/gateway/handlers/forwarding_proxy.go
@@ -133,7 +133,7 @@ func getServiceName(urlValue string) string {
 	if strings.HasPrefix(urlValue, forward) {
 		serviceName = urlValue[len(forward):]
 	}
-	return serviceName
+	return strings.Trim(serviceName, "/")
 }
 
 // LoggingNotifier notifies a log about a request
@@ -168,7 +168,6 @@ type FunctionAsHostBaseURLResolver struct {
 
 // Resolve the base URL for a request
 func (f FunctionAsHostBaseURLResolver) Resolve(r *http.Request) string {
-
 	svcName := getServiceName(r.URL.Path)
 
 	const watchdogPort = 8080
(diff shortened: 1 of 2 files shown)


LucasRoesler:
Note that a ticket should be opened in the openfaas-operatore because this same snippet of code is used there too (@stefanprodan)

LucasRoesler:
I have now tested this in a GKE cluster with the echo function.  The invoke works with `POST` and `GET`, with and without a trailing slash, and with `GET` parameters.  Here are the results via `curl`

```
$ curl -X POST http://gateway.openfaas.local/function/echo/ -d "test"
test
$ curl -X GET http://gateway.openfaas.local/function/echo/ -d "test"
test%
$ curl -X POST http://gateway.openfaas.local/function/echo/\?name\=foo -d "test"
test
$ curl -X GET http://gateway.openfaas.local/function/echo/\?name\=foo -d "test"
test
$ curl -X POST http://gateway.openfaas.local/function/echo -d "test"
test
$ curl -X GET http://gateway.openfaas.local/function/echo -d "test"
test
$ curl -X POST http://gateway.openfaas.local/function/echo\?name\=foo -d "test"
test
$ curl -X GET http://gateway.openfaas.local/function/echo\?name\=foo -d "test"
test
```

alexellis:
Thanks for this 👍 💯 for unit tests.

Do we need similar changes anywhere else? Thoughts are: queue worker, the connectors for Kafka / RabbitMQ and the rest of the gateway code? I thought this function may have been duplicated in 2 places.

Issue #590, opened 2018-03-20
Feature: read-only root file-system for containers


alexellis:
## Description

A container image is immutable, but once deployed as a running container changes can be made to the file-system aka the "root file-system" mounted at `/`. Relates to #144 

This feature implements a read-only filesystem to protect the function code and libraries from changes at run-time. I was told by @justincormack that cloud providers such as AWS Lambda do not support this use-case yet (it may have changed since).

Pros:
- enhanced security
- durable

Cons:
- could break some existing functions
- needs an additional temporary filesystem to create valid writable area

As a PoC I created a small patch for faas-netes which made all functions have a read-only file-system. This also involved mounting what Kubernetes calls an "empty-dir" to the `/tmp` mount-point.

Functions can use `/tmp` as a temporary storage area.

## Scope

* Any change should be supported in Kubernetes and Docker Swarm.
* Be configurable at a per-function level via the deployment API
* Be configurable via CLI YAML

LucasRoesler:
Copying from my comment on #144 

Couple of questions about how this is used:

1. What the practical use case where a developer explicitly opts into a read-only fs or would it be opt-out to a non-root fs?
2. With the above question in mind, do we see the openfaas cluster admin having an option to enforce read-only root functions?
3. If the cluster admin can control this, does it simply override the setting and return a warning if the function is not labelled as supporting a root-only fs?  In this case, the function would just break as soon as it tries to write to the fs.  Or, would it just outright reject it until it is labeled correctly?
4. Do we want to set a specific convention about having a writable folder, say `/opt/sandbox`?  Perhaps it is an ENV var that openfaas can pass in, `FAAS_WRITE_DIR`?

I will think a little more and come up with some more questions.

alexellis:
The writable folder would be `/tmp/` as specified. Kubernetes does not allow a hard limit on `empty_dir` mounts right now but @stefanprodan tells me it's coming soon.

Version 1 of this feature will be opt-in only through the YAML file / config.

Do you have enough information to start working on this?

alexellis:
Derek add label: help wanted

milsonian:
[...]

PR #723, opened 2018-06-16
[WIP] Adding support for optional read-only fs for functions

Templum:
This pull request lays the foundation for adding read-only fs functions. The flag needs to be set to true in order to enable it. By default, it will be disabled and behave like currently. 

## Motivation and Context
- [x] I have raised an issue to propose this change (#590)

## How Has This Been Tested?
The tests will be in the OpenFaaS provider.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/requests/requests.go
+++ b/gateway/requests/requests.go
@@ -41,6 +41,9 @@ type CreateFunctionRequest struct {
 
 	// Requests of resources requested by function
 	Requests *FunctionResources `json:"requests"`
+
+	// Flag for enabling a readonly filesystem for the function.
+	ReadOnlyFs bool `json:"readOnlyFs"`
 }
 
 // FunctionResources Memory and CPU


alexellis:
Simon, good to see this work starting. What is the flag called in the K8s and Swarm APIs - I'm curious about the casing they used too?

Alex

alexellis:
I believe the flag is `ReadOnlyRootFilesystem` in Kubernetes having check the client API code and [the docs](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) - I think this would be a more consistent choice. What do you think Simon?

Templum:
Hi @alexellis ,

I choose the name on purpose. In docker/swarm it is called `read_only` and in K8S like you already told it is called `ReadOnlyRootFilesystem `. And in that regards, I think the addition  `root filesystem` might irritate people. Also we want to support both plattforms, so I would recommend against reusing the wording and instead suggested  `ReadOnlyFs ` as this is straight forward and should also elimante any space for potential missunderstanding. 

Simon

alexellis:
> For K8S I could see people getting irritated by the addition root filesystem

This is the current naming in Kubernetes, what do you mean by K8s users being annoyed by the naming?

If we use `ReadOnlyRootFilesystem` instead of `ReadOnlyFs` then it shows that the root-filesystem specifically is going to be read-only. Does `ReadOnlyRoot` work any better?

What's the main concern in your opinion.. is it typing more letters?

Templum:
> This is the current naming in Kubernetes, what do you mean by K8s users being annoyed by the naming?

First of all my mistake there, I meant the docker user. As I have shown the naming is for them different. 

> If we use ReadOnlyRootFilesystem instead of ReadOnlyFs then it shows that the root-filesystem specifically is going to be read-only. Does ReadOnlyRoot work any better?

In that term as a docker user, you could assume that only the root part of the system is read-only, but what is with the userspace. I could technically still download software there, but as far as I saw (on Docker & K8s). The entire file system is read-only, @alexellis  can you follow my reasoning now? Other then this, it's just a name, I can also change it.

alexellis:
@stefanprodan please can you comment on the naming choice?

alexellis:
> The entire file system is read-only

Not quite right, we will need to give a /tmp/ mount for file PIDs and temporary files.

LucasRoesler:
I think either name will require documentation for people to use correctly, so I don't have a super strong opinion one way or the other.  But I think I would probably advocate for `ReadOnlyRootFS` 
 or `ReadOnlyRootFilesystem` because it seems that both K8S and AWS are using it in the same way, there may be others as well.

- AWS https://docs.aws.amazon.com/batch/latest/APIReference/API_ContainerProperties.html
- K8S https://kubernetes.io/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems

It is also more specific to the container use case when I do a google search for the term, even _this_ issue comes up along with docs.  If you search `ReadOnlyFS`, there are a lot of other unrelated results.  So, `ReadOnlyRootFS` is probably more discoverable and helpful for people. 

As a small nitpick,  I would advocate that if we abbreviate `filesystem`, then `fs` be capitalized to match the golint rules about initials (https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

alexellis:
I know it's a verbose option `ReadOnlyRootFilesystem`, but I feel matching the Kubernetes API will go in our favour since most of our users will be on Kubernetes and this will be one less mental mapping for them when using OpenFaaS.

Templum:
Then I will update the flag to match `ReadOnlyRootFilesystem`.

alexellis:
Thanks. I'll then merge it as soon as it's ready. Can you revendor it in the faas-cli after that?

alexellis:
@templum please can you repush today? We're blocked on the other work right now.

Thanks

PR #738, opened 2018-06-28
Updated default secret mount path value

viveksyngh:
This changes updates the secret mount path default value to
/var/openfaas/secrets at serveral places. It has been changed
from /run/secrets/ to /var/openfaas/secrets after version 0.8.2

## Motivation and Context

## How Has This Been Tested?

I have tested this on my local system.

## Types of changes

## Checklist:

### DIFF
--- a/docker-compose.armhf.yml
+++ b/docker-compose.armhf.yml
@@ -16,7 +16,7 @@ services:
             direct_functions: "true"    # Functions are invoked directly over the overlay network
             direct_functions_suffix: ""
             basic_auth: "false"
-            secret_mount_path: "/run/secrets/"
+            secret_mount_path: "/var/openfaas/secrets/"
         deploy:
             resources:
                 # limits:   # uncomment to enable limits
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,7 @@ services:
             direct_functions: "true"    # Functions are invoked directly over the overlay network
             direct_functions_suffix: ""
             basic_auth: "false"
-            secret_mount_path: "/run/secrets/"
+            secret_mount_path: "/var/openfaas/secrets/"
         deploy:
             resources:
                 # limits:   # Enable if you want to limit memory usage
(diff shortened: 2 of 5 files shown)


alexellis:
LGTM and CI passes too, thank you for picking this up @viveksyngh - really good to see the range of work you've taken on, it makes a difference to the project for sure. 👍

johnmccabe:
@viveksyngh / @alexellis how was the swarm `/run/secrets` mount point overridden when creating the secret during testing? The difference between Swarm/K8s config was surely down to the inability to specify a secret mount point in Swarm?

alexellis:
I assumed Vivek tested both. Actually this should have remained as it was for Swarm. The mount path here is not for a function so the new path is invalid unless someone specifies an override when creating their basic with secrets. 

Maybe it needs reverting.

johnmccabe:
I think so, there is apparently the ability to override the location using `--template-driver` since 17.06 but its poorly documented.

alexellis:
I won’t revert it via GitHub due to issues with sign-off. Can we get a follow up PR?

viveksyngh:
@alexellis This is something I think I missed in my previous PR. But What I did to modify mount path is changed the `docker-compose.yml` to this.

```
secrets:
            - source: basic-auth-user
              target: /var/openfaas/secrets/basic-auth-user
            - source: basic-auth-password
              target: /var/openfaas/secrets/basic-auth-password
```

viveksyngh:
If this is fine. We can update our `docker-compose.yml` and add documentation that if you are modifying secrets path, you need to update that here as well.

viveksyngh:
You just create the secret with usual `docker secret create` command. Update the secret definition for `gateway` to use `source` and `target` as below

```yaml
secrets:
            - source: basic-auth-user
              target: /var/openfaas/secrets/basic-auth-user
            - source: basic-auth-password
              target: /var/openfaas/secrets/basic-auth-password
```
When you will deploy the stack, It will mount to that path.

johnmccabe:
Was just about to delete that question after seeing your previous comment.

It'll be worth explaining how the change was tested in the fix PR, makes it less likely stuff slips through accidentally.

johnmccabe:
@alexellis do you care about versions older than 17.06 being supported? That'd be the only concern with using the target option.

viveksyngh:
Yeah, Agree! I am sorry about that.

viveksyngh:
@alexellis Should we update `stack.yml` file to mount the basic auth secrets to `/var/openfaas/secrets` using `target` or we revert the changes?

alexellis:
Reverting fixes the issue and won't need additional testing. Let's review in time whether to move to target syntax

Issue #74, opened 2017-05-01
Get to 1.0 - Security strategy


alexellis:
Initial areas for security strategy. Implementation needs to be easy to use when deployed locally or in development environment.

**UI**

Static Angular UI tester visible when deployed

Suggestion - HTTPs / basic auth and cookie. Minimum initial version could have a single admin account.

**System API**

Endpoints: deploy function / delete function / get a list of functions

Approach needs to be consumed by the UI and by an optional CLI.

**Functions API**

I.e. /function/function_name [POST]

This can be split into two use-cases:

For the functions expecting webhooks, it's rare that I've seen anything here - sometimes a digest of the message from the remote party. Mainly just a HTTP POST with a body to your endpoint.

For endpoints not receiving webhooks from third-parties, but being used as part of a chain of functions etc - maybe here it makes sense to add some kind of header/bearer token / JWT.

**Related**

* Rate limiting
* IP black/whitelisting
* Account lock-out for bad login attempts
* Account reset (via? email?)
* Is persistent storage needed for any of the above?

diogomonica:
In general seems like you're thinking about all the right things. I would say that supporting an HMAC injected as a header from any function receiving a web hook is probably the right call. That means that you probably need to have per-function symmetric keys?

Omeryl:
Any further thoughts on this in the long run?

alexellis:
Hi @Omeryl - I would say that this is outside the scope of a serverless framework and could be implemented easily with a project specialising in these areas such as NGinx / Traefik or Kong. Does that make sense? What are your thoughts?

Omeryl:
@alexellis I mean, as long as the client is capable of doing basic auth, or sending specific headers, it could definitely be built as plugins for the client. I was just wondering what the official thoughts were on this before I started tinkering with this for some coworkers is all.

alexellis:
Can you give me a quick outline of what you had in mind?

alexellis:
Kong/Mashape are working on a plugin for OpenFaaS - this should add whatever we need in terms of security / OAuth / basic auth / rate-limiting etc.

PR #743, opened 2018-07-03
Enable basic-auth by default

alexellis:
## Description

Enable basic-auth by default.

This changes ./deploy_stack.sh to generate and then apply secrets turning
on the built-in basic auth feature of the gateway.

## Motivation and Context

#350 #74 

The deployment script will enable basic-auth by default to help
avoid people deploying to a public IP with no protection from
malicious actors.

## How Has This Been Tested?

Tested on MacOS with and without --no-auth flag.
Does not apply for armhf or powershell.

Example workflow for testing:

screen shot 2018-07-03 at 10 25 22 am

Test with positive and negative workflows: positive - password supplied and consumed, show that auth is working by using wrong password. Negative flow - disable auth and check auth is not required.

## Types of changes

## Checklist:

Will need documentation change and change to integration tests to disable auth via `--no-auth` flag when deploying during testing.

### DIFF
--- a/deploy_stack.sh
+++ b/deploy_stack.sh
@@ -5,6 +5,50 @@ if ! [ -x "$(command -v docker)" ]; then
   exit 1
 fi
 
-echo "Deploying stack"
-docker stack deploy func --compose-file docker-compose.yml
+export BASIC_AUTH="true"
 
+sha_cmd="shasum -a 256"
+if ! command -v shasum >/dev/null; then
+  sha_cmd="sha256sum"
+fi
+
+while [ ! $# -eq 0 ]
+do
+	case "$1" in
+		--no-auth | -n)
+			export BASIC_AUTH="false"
+			;;
+    --help | -h)
+			echo "Usage: \n [default]\tdeploy the OpenFaaS core services\n --no-auth [-n]\tdisable basic authentication.\n --help\tdisplays this screen"
+      exit
+			;;
+	esac
+	shift
+done
+
+# Secrets should be created even if basic-auth is disabled.
+echo "Attempting to create credentials for gateway.."
+echo "admin" | docker secret create basic-auth-user -
+secret=$(head -c 16 /dev/urandom| $sha_cmd | cut -d " " -f 1)
+echo "$secret" | docker secret create basic-auth-password -
+if [ $? = 0 ];
+then
+  echo "[Credentials]\n username: admin \n password: $secret\n echo -n "$secret" | faas-cli login --username=admin --password-stdin"
+else
+  echo "[Credentials]\n already exist, not creating"
+fi
+
+if [ $BASIC_AUTH = "true" ];
+then
+  echo ""
+  echo "Enabling basic authentication for gateway.."
+  echo ""
+else
+  echo ""
+  echo "Disabling basic authentication for gateway.."
+  echo ""
+fi
+
+echo "Deploying OpenFaaS core services"
+
+docker stack deploy func --compose-file docker-compose.yml
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,7 @@ services:
             faas_nats_port: 4222
             direct_functions: "true"    # Functions are invoked directly over the overlay network
             direct_functions_suffix: ""
-            basic_auth: "false"
+            basic_auth: "${BASIC_AUTH:-true}"
             secret_mount_path: "/run/secrets/"
         deploy:
             resources:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ services:
     gateway:
         ports:
             - 8080:8080
-        image: openfaas/gateway:0.8.3
+        image: openfaas/gateway:0.8.5
         networks:
             - functions
         environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,9 +32,9 @@ services:
             placement:
                 constraints:
                     - 'node.platform.os == linux'
-        # secrets:
-        #     - basic-auth-user
-        #     - basic-auth-password
+        secrets:
+            - basic-auth-user
+            - basic-auth-password
 
     # Docker Swarm provider
     faas-swarm:
@@ -180,8 +180,8 @@ networks:
         labels:
           - "openfaas=true"
 
-# secrets:
-#     basic-auth-user:
-#         external: true
-#     basic-auth-password:
-#         external: true
+secrets:
+    basic-auth-user:
+        external: true
+    basic-auth-password:
+        external: true


mrtind:
When i test the change on docker playground deploy stack gives me 
./deploy_stack.sh: line 27: shasum: not found
and the password field is left blank.

alexellis:
That's good to know (probably because they use Alpine as a shell). Did the updated commit I pushed fix it? It seemed to fix it for me.

mrtind:
Now it works as expected.

alexellis:
Great 👍 

I saw some issues on DO with it hanging when I typed "head -c 1 /dev/random" - wonder if this is an issue with using VMs?

viveksyngh:
This changes looks fine and works as expected. 

But one thing I noticed, when we deploy stack with `--no-auth` flag it deploys and changes the `docker-compose.yml`. When we again try to deploy the stack with auth it does not change it back and deploys with `basic_auth: false`. It took me some time to figure out, I was wondering the auth is not working. I think we should add that check.

deploy_stack.sh
```diff
@@ -5,6 +5,50 @@ if ! [ -x "$(command -v docker)" ]; then
...
+  echo "[Credentials]\n username: admin \n password: $secret\n echo -n "$secret" | faas-cli login --username=admin --password-stdin"
+else
+  echo "[Credentials]\n already exist, not creating"
+fi
 
+if [ $auth -eq "1" ];
+then
+  echo ""
+  echo "Enabling basic authentication for gateway.."
+  echo ""
```
johnmccabe: I think you need to include the `sed` command here as well.

Say for example you deploy a stack with `--no-auth` this will alter the `docker-compose.yml` disabling auth, but if you then tear down the stack and run again without disabling auth, then you'd get the same altered compose file so auth would remain disabled.

deploy_stack.sh
```diff
@@ -5,6 +5,50 @@ if ! [ -x "$(command -v docker)" ]; then
...
+else
+  echo ""
+  echo "Disabling basic authentication for gateway.."
+  echo ""
+  sed -i -r.bak 's/basic_auth: \"true\"/basic_auth: \"false\"/' docker-compose.yml
+fi
+
+echo "Deploying OpenFaaS core services"
```
johnmccabe: For consistency you'd want the ellipsis at the end of this output.

deploy_stack.sh
johnmccabe: What about using envvars in the compose file instead (https://docs.docker.com/compose/environment-variables/) set enabled=true/false in these conditionals then run the final deploy with:
```
OPENFAAS_AUTH=$enabled docker stack deploy func --compose-file docker-compose.yml
```

deploy_stack.sh
```diff
@@ -5,6 +5,50 @@ if ! [ -x "$(command -v docker)" ]; then
...
+# Secrets should be created even if basic-auth is disabled.
+echo "Attempting to create credentials for gateway.."
+echo "admin" | docker secret create basic-auth-user -
+secret=$(head -c 16 /dev/urandom| $sha_cmd | cut -d " " -f 1)
+echo "$secret" | docker secret create basic-auth-password -
+if [ $? = 0 ];
+then
+  echo "[Credentials]\n username: admin \n password: $secret\n echo -n "$secret" | faas-cli login --username=admin --password-stdin"
+else
+  echo "[Credentials]\n already exist, not creating"
```
johnmccabe: I'd be inclined to give the user the option to replace these secrets. Could be in a situation where they've been setup in the past but the user is no longer sure what they had been set to (*cough* me).

deploy_stack.sh
```diff
@@ -5,6 +5,50 @@ if ! [ -x "$(command -v docker)" ]; then
...
+done
+
+# Secrets should be created even if basic-auth is disabled.
+echo "Attempting to create credentials for gateway.."
+echo "admin" | docker secret create basic-auth-user -
+secret=$(head -c 16 /dev/urandom| $sha_cmd | cut -d " " -f 1)
+echo "$secret" | docker secret create basic-auth-password -
+if [ $? = 0 ];
+then
+  echo "[Credentials]\n username: admin \n password: $secret\n echo -n "$secret" | faas-cli login --username=admin --password-stdin"
```
johnmccabe: The `echo -n...` needs some context rather than just printing along with the username and password. Ie blank line spacer and some text telling the user to use this to log into the gateway with the client.

Or even better, offer to do so on behalf of the user (if faas-cli is found on the path).

johnmccabe: You should also make very clear that the user should record/store the generated user credentials as they'll not be made available again.

alexellis:
@viveksyngh could you try modifying locally with the env-var variation suggestion by John? If that looks good I'll add it too.

viveksyngh:
@alexellis I will try that.

viveksyngh:
I will have this one ready over the weekend.

viveksyngh:
@alexellis I have added it here and It works as I have tested on my local.

https://github.com/viveksyngh/faas/commit/51f7818dcc7fb17bcac5f41e595af5babcdd47ec

deploy_stack.sh
```diff
@@ -5,6 +5,50 @@ if ! [ -x "$(command -v docker)" ]; then
...
+      exit
+			;;
+	esac
+	shift
+done
+
+# Secrets should be created even if basic-auth is disabled.
+echo "Attempting to create credentials for gateway.."
+echo "admin" | docker secret create basic-auth-user -
+secret=$(head -c 16 /dev/urandom| $sha_cmd | cut -d " " -f 1)
```
ivanayov: Would be good if there is an easy way for the user to create own credentials or edit the password without the need to set-up secrets manually.

What about passing and argument to `deploy_stack.sh`?

deploy_stack.sh
alexellis: I prefer the env-var approach with a default set.

deploy_stack.sh
alexellis: How about `docker secret create`?

deploy_stack.sh
alexellis: What context should we give about `-n`?

deploy_stack.sh
alexellis: Possibly in a follow-up PR. I think it's out of scope for this change. The user really must run the `faas-cli login` command though.

deploy_stack.sh
alexellis: How about something like this in the troubleshooting guide?

```
docker service create --image alpine:3.7 --secret secret-name cat /run/secrets/secret-name 
```

alexellis:
Derek set milestone: 0.8.6

alexellis:
Somehow the PR I merged from @viveksyngh brought in way too many commits and I couldn't resolve the issues. I've now rebased and force-pushed with a longer commit message and give credit to @viveksyngh for the part he added.

This is a breaking change so please can: @ivanayov @mrtind  @johnmccabe take another look?

Alex

alexellis:
Docs PR: https://github.com/openfaas/docs/pull/45

mrtind:
Ran new OpenFaas with auto-auth branch on a cloud deployed a function it seems that everything works.

alexellis:
Thanks @mrtind - did you do a "negative test" too? i.e. deploy with `--no-auth`?

rgee0:
Haven't got a great deal of time right now, but since the change 2 days ago I'm seeing deployment generate a password but not apply it - the UI is open.

I'll have a closer look when I can. Logging here in case someone else can have a look in the meantime.

viveksyngh:
@rgee0 can you check if your `docker-compose.yml` was modified when you tried to deploy it with `deploy_stack.sh  --no-auth`. It happened to me as well.  We have fixed that using env-var for BASIC_AUTH. Can you also take a pull and try again?

rgee0:
@viveksyngh I'm working on the automation & pull clean on each run.  It worked last week and it doesn't now.  Has the mode of operation changed?

rgee0:
I think I've found it.  In `docker-compose.yml` we see `basic_auth: "false"` and there is nothing in `deploy_stack.sh` to alter this.

viveksyngh:
Yes, that's what I was talking about. (y)

alexellis:
The work from @viveksyngh was tested.. so how can this be broken @rgee0 ? The env-var is set to true on line 8.

alexellis:
I've restored the changes but can't get the login to work via the `faas-cli` now.. @rgee0 @viveksyngh

alexellis:
All fixed now and re-tested with Swarm. The issue was a local stack.yml in the faas folder with a gateway URL not found in the openfaas config file. Maybe we can bubble that up better as an error?

rgee0:
Glad we caught that.  Looks a lot better through my automation now.

PR #767, opened 2018-07-13
Update .DEREK.yml with username

ivanayov:
Change `iyovcheva` to `ivanayov` after username has been changed

Signed-off-by: Ivana Yovcheva (VMware) <iyovcheva@vmware.com>


alexellis:
Derek says thanks 🤖

Issue #716, opened 2018-06-06
Proposal: Allow for path parameters by passing full URI to of-watchdog HTTP functions


telackey:
The new of-watchdog HTTP-style functions (cf. https://github.com/openfaas-incubator/node8-express-template) have a long-running HTTP server behind the actual function code.  This is very similar to other FaaS solutions, such as AWS Lambda, and is mostly hidden from the function developer.

The end result offers greater flexibility as arbitrary headers, response codes, etc. can be more easily returned.  It also eases the porting effort from other FaaS platforms.

One limitation of the current implementation is that only the query string of the incoming request, not its path, is passed along to the function.  Both the API Gateway and of-watchdog reconstruct the URL without the full path.  This means that I can make use of a request like:
  `/function/myfunc?category=123`

but not:
  `/function/myfunc/category/123`

## Desired Behaviour

Pass the complete URI all the way to the function, so that the function developer can decide which is most appropriate for their needs: path parameters or query string parameters.

## Current Behaviour
The full path is swallowed by the Gateway and by of-watchdog

## Possible Solution
I have attached some very simple, proof-of-concept patches (against 0.8.1) showing a possible solution.

[faas.diff.txt](https://github.com/openfaas/faas/files/2077881/faas.diff.txt)
[of-watchdog.diff.txt](https://github.com/openfaas/faas/files/2077882/of-watchdog.diff.txt)

iyovcheva:
This looks like a good-to-have.
@alexellis @rgee0 @ericstoekl any thoughts?

iyovcheva:
Derek add label: proposal

gmcelhoe:
This is very exciting, I was trying to write a REST-like API using OpenFaas and was getting stuck.  I'll try applying these changes locally.

alexellis:
@gmcelhoe what is the most useful part of the change for you?

For me it's being able to store a list of "topics" in a label/annotation for use with our event brokers such as the Kafka broker. Right now we can't store arbitrary data or even a comma in the strict Kubernetes format.

i.e. 

Currently valid:

```
  name: audit-function
  labels:
    topic: "kafka.payment.received"
```

Currently invalid:

```
  name: audit-function
  labels:
    topic: "kafka_payment_received"
```


Currently invalid:

```
  name: audit-function
  labels:
    topic: "topic1,topic2,topic3,topic4"
```

alexellis:
[...]

PR #789, opened 2018-07-25
 Implement proposal 716, passing full URL paths

telackey:
An implementation of proposal #716, passing the full URL through the Gateway and watchdog.

## Description

Previous to these changes, only the query string provide by the HTTP client was passed through to the handling code.  With these changes, the full URL path is passed as well as the query string.

## Motivation and Context
- [X] I have raised an issue to propose this change

Proposal #716 

## How Has This Been Tested?

A new automated test case has been added for the Gateway and for fwatchdog as well as manual testing.

## Types of changes

It is unlikely this is a breaking change.  Since the path is not currently passed by the Gateway, no existing function code can be relying on it.  Or more correctly, for there to be a problem, the code would have to be relying on the _absence_ of the path, which is unlikely though not completely impossible.

## Checklist:

This may or may not require a documentation change.  The existing documentation, AFAIK, does not address the use of paths one way or the other.

## Refs

Trello: https://trello.com/c/1qZMlGXU

### DIFF
--- a/gateway/handlers/forwarding_proxy.go
+++ b/gateway/handlers/forwarding_proxy.go
@@ -29,12 +29,17 @@ type BaseURLResolver interface {
 	Resolve(r *http.Request) string
 }
 
+// RequestURLPathTransformer Transform the incoming URL path for upstream requests
+type URLPathTransformer interface {
+	Transform(r *http.Request) string
+}
+
 // MakeForwardingProxyHandler create a handler which forwards HTTP requests
-func MakeForwardingProxyHandler(proxy *types.HTTPClientReverseProxy, notifiers []HTTPNotifier, baseURLResolver BaseURLResolver) http.HandlerFunc {
+func MakeForwardingProxyHandler(proxy *types.HTTPClientReverseProxy, notifiers []HTTPNotifier, baseURLResolver BaseURLResolver, urlPathTransformer URLPathTransformer) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		baseURL := baseURLResolver.Resolve(r)
 
-		requestURL := r.URL.Path
+		requestURL := urlPathTransformer.Transform(r)
 
 		start := time.Now()
 
@@ -51,23 +56,7 @@ func MakeForwardingProxyHandler(proxy *types.HTTPClientReverseProxy, notifiers [
 }
 
 func buildUpstreamRequest(r *http.Request, baseURL string, requestURL string) *http.Request {
-	url := baseURL
-
-	if requestURL != "" {
-		// When forwarding to a function, since the `/function/xyz` portion
-		// of a path like `/function/xyz/rest/of/path` is only used or needed
-		// by the Gateway, we want to trim it down to `/rest/of/path` for the
-		// upstream request.  In the following regex, in the case of a match
-		// the requestURL will be at `0`, the function name at `1` and the
-		// rest of the path (the part we are interested in) at `2`.
-		matcher := functionMatcher.Copy()
-		parts := matcher.FindStringSubmatch(requestURL)
-		if 3 == len(parts) {
-			url += parts[2]
-		} else {
-			url += requestURL
-		}
-	}
+	url := baseURL + requestURL
 
 	if len(r.URL.RawQuery) > 0 {
 		url = fmt.Sprintf("%s?%s", url, r.URL.RawQuery)
@@ -210,4 +199,47 @@ func (f FunctionAsHostBaseURLResolver) Resolve(r *http.Request) string {
 	}
 
 	return fmt.Sprintf("http://%s%s:%d", svcName, suffix, watchdogPort)
+}
+
+// TransparentURLPathTransformer passes the requested URL path through untouched.
+type TransparentURLPathTransformer struct {
+}
+
+// Transform returns the URL path unchanged.
+func (f TransparentURLPathTransformer) Transform(r *http.Request) string {
+	return r.URL.Path
+}
+
+// PathTruncatingURLPathTransformer always truncated the path to "/".
+type PathTruncatingURLPathTransformer struct {
+}
+
+// Transform always return a path of "/".
+func (f PathTruncatingURLPathTransformer) Transform(r *http.Request) string {
+	return "/"
+}
+
+// FunctionPrefixTrimmingURLPathTransformer removes the "/function/servicename/" prefix from the URL path.
+type FunctionPrefixTrimmingURLPathTransformer struct {
+}
+
+// Transform removes the "/function/servicename/" prefix from the URL path.
+func (f FunctionPrefixTrimmingURLPathTransformer) Transform(r *http.Request) string {
+	ret := r.URL.Path
+
+	if ret != "" {
+		// When forwarding to a function, since the `/function/xyz` portion
+		// of a path like `/function/xyz/rest/of/path` is only used or needed
+		// by the Gateway, we want to trim it down to `/rest/of/path` for the
+		// upstream request.  In the following regex, in the case of a match
+		// the r.URL.Path will be at `0`, the function name at `1` and the
+		// rest of the path (the part we are interested in) at `2`.
+		matcher := functionMatcher.Copy()
+		parts := matcher.FindStringSubmatch(ret)
+		if 3 == len(parts) {
+			ret = parts[2]
+		}
+	}
+
+	return ret
 }
--- a/gateway/handlers/forwarding_proxy.go
+++ b/gateway/handlers/forwarding_proxy.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -14,6 +15,9 @@ import (
 	"github.com/openfaas/faas/gateway/types"
 	"github.com/prometheus/client_golang/prometheus"
 )
+
+// Parse out the service name (group 1) and rest of path (group 2).
+var functionMatcher = regexp.MustCompile("^/?function/([^/?]+)([^?]*)")
 
 // HTTPNotifier notify about HTTP request/response
 type HTTPNotifier interface {
@@ -46,7 +50,24 @@ func MakeForwardingProxyHandler(proxy *types.HTTPClientReverseProxy, notifiers [
 	}
 }
 
-func buildUpstreamRequest(r *http.Request, url string) *http.Request {
+func buildUpstreamRequest(r *http.Request, baseURL string, requestURL string) *http.Request {
+	url := baseURL
+
+	if requestURL != "" {
+		// When forwarding to a function, since the `/function/xyz` portion
+		// of a path like `/function/xyz/rest/of/path` is only used or needed
+		// by the Gateway, we want to trim it down to `/rest/of/path` for the
+		// upstream request.  In the following regex, in the case of a match
+		// the requestURL will be at `0`, the function name at `1` and the
+		// rest of the path (the part we are interested in) at `2`.
+		matcher := functionMatcher.Copy()
+		parts := matcher.FindStringSubmatch(requestURL)
+		if 3 == len(parts) {
+			url += parts[2]
+		} else {
+			url += requestURL
+		}
+	}
 
 	if len(r.URL.RawQuery) > 0 {
 		url = fmt.Sprintf("%s?%s", url, r.URL.RawQuery)
@@ -69,7 +90,7 @@ func buildUpstreamRequest(r *http.Request, url string) *http.Request {
 
 func forwardRequest(w http.ResponseWriter, r *http.Request, proxyClient *http.Client, baseURL string, requestURL string, timeout time.Duration) (int, error) {
 
-	upstreamReq := buildUpstreamRequest(r, baseURL+requestURL)
+	upstreamReq := buildUpstreamRequest(r, baseURL, requestURL)
 	if upstreamReq.Body != nil {
 		defer upstreamReq.Body.Close()
 	}
@@ -134,7 +155,16 @@ func getServiceName(urlValue string) string {
 	var serviceName string
 	forward := "/function/"
 	if strings.HasPrefix(urlValue, forward) {
-		serviceName = urlValue[len(forward):]
+		// With a path like `/function/xyz/rest/of/path?q=a`, the service
+		// name we wish to locate is just the `xyz` portion.  With a postive
+		// match on the regex below, it will return a three-element slice.
+		// The item at index `0` is the same as `urlValue`, at `1`
+		// will be the service name we need, and at `2` the rest of the path.
+		matcher := functionMatcher.Copy()
+		matches := matcher.FindStringSubmatch(urlValue)
+		if 3 == len(matches) {
+			serviceName = matches[1]
+		}
 	}
 	return strings.Trim(serviceName, "/")
 }
--- a/gateway/handlers/forwarding_proxy.go
+++ b/gateway/handlers/forwarding_proxy.go
@@ -210,13 +210,26 @@ func (f TransparentURLPathTransformer) Transform(r *http.Request) string {
 	return r.URL.Path
 }
 
-// PathTruncatingURLPathTransformer always truncated the path to "/".
-type PathTruncatingURLPathTransformer struct {
+// FunctionPathTruncatingURLPathTransformer always truncated the path to "/".
+type FunctionPathTruncatingURLPathTransformer struct {
 }
 
 // Transform always return a path of "/".
-func (f PathTruncatingURLPathTransformer) Transform(r *http.Request) string {
-	return "/"
+func (f FunctionPathTruncatingURLPathTransformer) Transform(r *http.Request) string {
+	ret := r.URL.Path
+
+	if ret != "" {
+		matcher := functionMatcher.Copy()
+		parts := matcher.FindStringSubmatch(ret)
+		// In the following regex, in the case of a match the r.URL.Path will be at `0`,
+		// the function name at `1` and the rest of the path (the part we are interested in)
+		// at `2`.  For this transformer, all we need to do is confirm it is a function.
+		if 3 == len(parts) {
+			ret = "/"
+		}
+	}
+
+	return ret
 }
 
 // FunctionPrefixTrimmingURLPathTransformer removes the "/function/servicename/" prefix from the URL path.
--- a/gateway/handlers/forwarding_proxy.go
+++ b/gateway/handlers/forwarding_proxy.go
@@ -17,7 +17,7 @@ import (
 )
 
 // Parse out the service name (group 1) and rest of path (group 2).
-var functionMatcher = regexp.MustCompile("^/?function/([^/?]+)([^?]*)")
+var functionMatcher = regexp.MustCompile("^/?(?:async-)?function/([^/?]+)([^?]*)")
 
 // HTTPNotifier notify about HTTP request/response
 type HTTPNotifier interface {
--- a/gateway/handlers/forwarding_proxy_test.go
+++ b/gateway/handlers/forwarding_proxy_test.go
@@ -20,7 +20,7 @@ func Test_buildUpstreamRequest_Body_Method_Query(t *testing.T) {
 		t.Fail()
 	}
 
-	upstream := buildUpstreamRequest(request, "/")
+	upstream := buildUpstreamRequest(request, "/", "")
 
 	if request.Method != upstream.Method {
 		t.Errorf("Method - want: %s, got: %s", request.Method, upstream.Method)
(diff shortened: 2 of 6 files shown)


gateway/handlers/forwarding_proxy.go
```diff
@@ -46,7 +47,18 @@ func MakeForwardingProxyHandler(proxy *types.HTTPClientReverseProxy, notifiers [
 	}
 }
 
-func buildUpstreamRequest(r *http.Request, url string) *http.Request {
+func buildUpstreamRequest(r *http.Request, baseURL string, requestURL string) *http.Request {
+	url := baseURL
+
+	if requestURL != "" {
+		matcher, _ := regexp.Compile("/?function/([^/?]+)([^?]*)")
```
alexellis: Hey, just curious what the regex is for?

gateway/handlers/forwarding_proxy.go
```diff
@@ -131,7 +143,12 @@ func getServiceName(urlValue string) string {
 	var serviceName string
 	forward := "/function/"
 	if strings.HasPrefix(urlValue, forward) {
-		serviceName = urlValue[len(forward):]
+		r, _ := regexp.Compile(forward + "([^/?]+).*")
+		matches := r.FindStringSubmatch(urlValue)
+		// The matched subgroup will be the function name.
+		if 2 == len(matches) {
```
alexellis: What does the magic number 2 mean? What is the purpose of the regex?

gateway/handlers/forwarding_proxy.go
telackey: Good question!  I allowed some of my regex related comments to fall by the wayside when I refactored.  I will add them back...

So, when building the upstream URL, when forwarding it for a function, it is helpful to strip off the `/function/{functionName}` prefix to the Gateway way.

That is, if I request:  `http://gateway:8080/function/xyz/my/path` it is nicer for it to forward to:  'http://xyzcontainer:8080/my/path` than `http://xyzcontainer:8080/function/xyz/my/path`.

This regex lifts out the URL after `/function/{functionName}` and appends it to the baseURL.

This should only be done for function forwarding though, as the Gateway also forwards other types of requests and shouldn't mangle those.

gateway/handlers/forwarding_proxy.go
telackey: Similar to the above on adding back some comments...

So in this case, we want to figure out he function name from the URL.  Previously that was done by trimming off the `forward` length and taking the rest.  With a case like: `/function/xyz/my/path` though, what we really want is the section just following `/function/` up to, but not including any later path segments or query string.

We've got one regex subgroup to match, so per the `FindStringSubmatch` if there is a match, the returned slice will have the text of the leftmost match at `matches[0]` (in this case, `urlValue`) followed by the matched subgroups.  We've only got one subgroup, so what we want is `matches[1]`.  So what we want to see for a positive match is a `matches` where the length is exactly `2`.

I'll try to come up with a briefer way to say all that for the comment.

ericstoekl:
@telackey Thanks for your PR. Could you give some steps that I can run to test out the functionality?

I tried building it this version of the gateway, and deployed a function which lists environment variables.

```
~/github/go/src/github.com/openfaas/faas/watchdog $ curl localhost:8080/function/pytest?hi=1
{'LANG': 'C.UTF-8', 'Http_Accept': '*/*', 'Http_Method': 'GET', 'GPG_KEY': 'C01E1CAD5EA2C4F0B8E3571504C367C218ADD4FF', 'PYTHON_VERSION': '2.7.14', 'Http_X_Start_Time': '1533009066872625442', 'Http_X_Forwarded_For': '10.255.0.2:49476', 'PYTHON_PIP_VERSION': '9.0.1', 'HOSTNAME': '040a8fa90a99', 'fprocess': 'python index.py', 'Http_Query': 'hi=1', 'Http_ContentLength': '0', 'Http_User_Agent': 'curl/7.47.0', 'Http_X_Call_Id': 'abfe5986-d2ef-4d22-b3c8-ee8f47ca8621', 'HOME': '/home/app', 'PATH': '/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/app/.local/bin', 'Http_Accept_Encoding': 'gzip', 'Http_Path': '/'}

~/github/go/src/github.com/openfaas/faas/watchdog $ curl localhost:8080/function/pytest/hi
{'LANG': 'C.UTF-8', 'Http_Accept': '*/*', 'Http_Method': 'GET', 'GPG_KEY': 'C01E1CAD5EA2C4F0B8E3571504C367C218ADD4FF', 'PYTHON_VERSION': '2.7.14', 'Http_X_Start_Time': '1533009075455684100', 'Http_X_Forwarded_For': '10.255.0.2:49680', 'PYTHON_PIP_VERSION': '9.0.1', 'HOSTNAME': '040a8fa90a99', 'fprocess': 'python index.py', 'Http_ContentLength': '0', 'Http_User_Agent': 'curl/7.47.0', 'Http_X_Call_Id': '0b2de2bd-183d-419d-8e14-fd7bc248104b', 'HOME': '/home/app', 'PATH': '/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/app/.local/bin', 'Http_Accept_Encoding': 'gzip', 'Http_Path': '/hi'}
```

In the second call, should `hi` show up in the `Http_Query` list? Or would you expect it to just show up in the path?

telackey:
@ericstoekl 

Just on the `Http_Path`.  If you try it with the old code, you should find that you get a path of `/` with both of your examples.

It really comes into its own when combined with something like the new node8 Express template:  https://github.com/openfaas-incubator/node8-express-template/, where Express could be used to handle the routing.

telackey:
@ericstoekl 

Following up on the mention of the Node8 Express template, if you wanted to do some more extensive checking, you might use something like this in the `index.js`:

```
if (isObject(handler)) {
    _.forOwn(handler, function(value, key) {
        if (_.isFunction(value)) {
            if ('root' === key) {
                console.log("Dynamically registered route: /");
                app.get('/', value);
                app.post('/', value);
            } else {
                console.log("Dynamically registered route: /" +  key);
                app.post('/' + key, value);
                app.get('/' + key, value);
            }
        }
    });
} else {
    console.log("Registered catch-all route: /*");
    app.get('/*', handler);
    app.post('/*', handler);
}
```
This would allow you to do something like this in `handler.js`:

```
module.exports = {
     a: ... fnA ... , 
     b:  ... fnB ...,
     root: ... fnX ...
};
```

Which you could then test by hitting `/`, `/a`, `/b`, etc.  Without the patches (ie, this and the related `of-watchdog` patch), you will only hit the root handler regardless, but after the patches you should be able to hit each according to the path used.

gateway/handlers/forwarding_proxy.go
```diff
@@ -46,7 +47,24 @@ func MakeForwardingProxyHandler(proxy *types.HTTPClientReverseProxy, notifiers [
...
+	url := baseURL
+
+	if requestURL != "" {
+		// When forwarding to a function, since the `/function/xyz` portion
+		// of a path like `/function/xyz/rest/of/path` is only used or needed
+		// by the Gateway, we want to trim it down to `/rest/of/path` for the
+		// upstream request.  In the following regex, in the case of a match
+		// the requestURL will be at `0`, the function name at `1` and the
+		// rest of the path (the part we are interested in) at `2`.
+		matcher, _ := regexp.Compile("/?function/([^/?]+)([^?]*)")
```
alexellis: During review I noticed that this compilation could be made ahead of time and re-used. This should improve performance/throughput.

alexellis:
I deployed the image to my swarm cluster and function readiness appears to be broken too.

screen shot 2018-08-02 at 12 34 05

telackey:
@alexellis 

The readiness problem was an issue with the regex matching '/system/function/' as well as '/function'.  This is my own fault for changing the code up at the 11th hour (previously, all the trimming had been done in the `watchdog`.)

It should be fixed now.  As regards the IP issue... I cannot reproduce that, and it is hard to see how any of the code changes could be related.  I'd suggest re-trying on the latest code, and if you still see it, I'll make another attempt.

telackey:
@alexellis 

I also addressed your comment about the regex.  By my understanding, sharing regexs in golang can be tricky because you can end up with significant contention on the internal locks.  My fix here is to pre-compile the expression as you suggested, but use `Copy()` for the actual usage.  I also simplified to using a single expression for both the service name matching an the function path matching (it was pretty much the same expression anyway).

telackey:
That string of merges is just a battle against git and amending with a signoff.

alexellis:
This is high up on my list @telackey

gateway/handlers/forwarding_proxy.go
```diff
@@ -66,7 +87,7 @@ func buildUpstreamRequest(r *http.Request, url string) *http.Request {
 
 func forwardRequest(w http.ResponseWriter, r *http.Request, proxyClient *http.Client, baseURL string, requestURL string, timeout time.Duration) (int, error) {
 
-	upstreamReq := buildUpstreamRequest(r, baseURL+requestURL)
+	upstreamReq := buildUpstreamRequest(r, baseURL, requestURL)
```
alexellis: I had a previous question about whether this belongs in the generic code which forwards a request upstream. I feel like it belongs in a middleware handler which wraps this one, in that way we retain single-responsibility and the transformation is limited to the routes we add the middleware to. This would also allow one piece of code to cover the `/function/` and `/async-function/` routes.

alexellis: A good example might be the CallID middleware?

alexellis: Or the basic auth middleware

watchdog/requesthandler_test.go
```diff
@@ -430,6 +430,40 @@ func TestHealthHandler_SatusMethoNotAllowed_ForWriteableVerbs(t *testing.T) {
 	}
 }
 
+func TestHandler_HasFullPathAndQueryInFunction_WithCgi_Mode(t *testing.T) {
```
alexellis: Since this is not changing behaviour, just documenting it, I think we could raise a separate PR for it and I'll merge right aware since it's not changing anything.

alexellis:
@telackey I noticed you didn't respond to my feedback on the unit test, so I've gone ahead and completed that task for you.

Please can you remove the unit test in `watchdog/requesthandler_test.go` from your changeset and squash down all 6 commits then push back up?

@ivanayov is doing some end to end validation of your changes, but I'd like to be able to test them against master too.

Alex

telackey:
@alexellis Per your earlier request, I changed things up to use a new `URLPathTransformer` interface instead.

I also added a configuration option:  `pass_url_path_to_functions`.  When this is `true` the path following the '/function/xyz/` portion is passed along upstream.  When it is `false`, the behavior will be similar to what it is now, passing only `/` upstream.

ivanayov:
I've tested with single and multiple parameters. My function prints "Http_Path" and "Http_Query".

First thought there is a problem with multiple parameters, but the bug was in my code.
So it works pretty well in this case:
```
$ curl "http://127.0.0.1:8080/function/apitest?arg=test&arg2=tralala&arg3=shalala&arg4=justonemorea"
{"path": "/", "query": "arg=test&arg2=tralala&arg3=shalala&arg4=justonemorea"}

$ curl  http://127.0.0.1:8080/function/apitest/arg/test/arg2/tralala/arg3/shalala/arg4/justonemore
{"path": "/arg/test/arg2/tralala/arg3/shalala/arg4/justonemore", "query": null}
```

gateway/handlers/forwarding_proxy.go
```diff
@@ -15,6 +16,9 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+// Parse out the service name (group 1) and rest of path (group 2).
+var functionMatcher = regexp.MustCompile("^/?(?:async-)?function/([^/?]+)([^?]*)")
```
ivanayov: This needs considering whether `GET` should be allowed for asynchronous functions. Not sure it makes sense to be supported, but if path parameters work for both `function` and `async-function` the user may expect to be able to reference the same path with `async` as without.

gateway/handlers/forwarding_proxy.go
```diff
@@ -178,3 +197,59 @@ func (f FunctionAsHostBaseURLResolver) Resolve(r *http.Request) string {
...
+	if ret != "" {
+		// When forwarding to a function, since the `/function/xyz` portion
+		// of a path like `/function/xyz/rest/of/path` is only used or needed
+		// by the Gateway, we want to trim it down to `/rest/of/path` for the
+		// upstream request.  In the following regex, in the case of a match
+		// the r.URL.Path will be at `0`, the function name at `1` and the
+		// rest of the path (the part we are interested in) at `2`.
+		matcher := functionMatcher.Copy()
+		parts := matcher.FindStringSubmatch(ret)
+		if 3 == len(parts) {
```
alexellis: I think we're close now but I don't like the magic numbers. Can we make this clearer for the code maintainers?

gateway/queue/types.go
```diff
@@ -11,6 +11,7 @@ type Request struct {
 	Header      http.Header
 	Body        []byte
 	Method      string
+	Path        string
```
alexellis: 👍 do we have the equivalent change lined-up in nats-queue-worker @telackey @ivanayov ?

gateway/server.go
```diff
@@ -72,14 +72,23 @@ func main() {
...
+	urlTransformer := handlers.FunctionPathTruncatingURLPathTransformer{}
+	var functionURLTransformer handlers.URLPathTransformer
 
-	faasHandlers.RoutelessProxy = handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver)
-	faasHandlers.ListFunctions = handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver)
-	faasHandlers.DeployFunction = handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver)
-	faasHandlers.DeleteFunction = handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver)
-	faasHandlers.UpdateFunction = handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver)
-	queryFunction := handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver)
+	if config.PassURLPathsToFunctions {
```
alexellis: Maybe we should drop the config and enable this all the time? If it's not breaking

gateway/handlers/forwarding_proxy.go
alexellis: I think the async route only allows for `POST` right now, but we could extend that. Please @ivanayov check and raise an issue to update if that's the case.

alexellis:
FYI we also have a rebase issue showing up with `gateway/server.go`.

alexellis:
I'm keen to move this forward and see it merged.

gateway/handlers/forwarding_proxy.go
```diff
@@ -25,12 +29,17 @@ type BaseURLResolver interface {
...
 }
 
+// RequestURLPathTransformer Transform the incoming URL path for upstream requests
+type URLPathTransformer interface {
+	Transform(r *http.Request) string
+}
+
 // MakeForwardingProxyHandler create a handler which forwards HTTP requests
-func MakeForwardingProxyHandler(proxy *types.HTTPClientReverseProxy, notifiers []HTTPNotifier, baseURLResolver BaseURLResolver) http.HandlerFunc {
+func MakeForwardingProxyHandler(proxy *types.HTTPClientReverseProxy, notifiers []HTTPNotifier, baseURLResolver BaseURLResolver, urlPathTransformer URLPathTransformer) http.HandlerFunc {
```
ivanayov: Let's split on two lines, as this becomes hard for reviewing.

gateway/handlers/forwarding_proxy_test.go
```diff
@@ -116,3 +116,50 @@ func Test_getServiceName(t *testing.T) {
 		})
 	}
 }
+
+func Test_buildUpstreamRequest_Body_Method_Query_Path(t *testing.T) {
+	srcBytes := []byte("hello world")
+	path := "/my/deep/path"
+
+	reader := bytes.NewReader(srcBytes)
+	request, _ := http.NewRequest(http.MethodPost, "/function/xyz"+path+"?code=1", reader)
```
ivanayov: Should have a test with `async-function` as well

ivanayov:
The branch needs rebase from master.

gateway/handlers/forwarding_proxy.go
```diff
@@ -178,3 +197,59 @@ func (f FunctionAsHostBaseURLResolver) Resolve(r *http.Request) string {
...
+func (f FunctionPathTruncatingURLPathTransformer) Transform(r *http.Request) string {
+	ret := r.URL.Path
+
+	if ret != "" {
+		matcher := functionMatcher.Copy()
+		parts := matcher.FindStringSubmatch(ret)
+		// In the following regex, in the case of a match the r.URL.Path will be at `0`,
+		// the function name at `1` and the rest of the path (the part we are interested in)
+		// at `2`.  For this transformer, all we need to do is confirm it is a function.
+		if 3 == len(parts) {
```
ivanayov: Similar code is repeated on 3 places returning different result. Would be better to go in a single method to make the easier support in future.

gateway/queue/types.go
telackey: No, I am not in a good position to test async functionality.

gateway/handlers/forwarding_proxy.go
```diff
@@ -131,7 +141,16 @@ func getServiceName(urlValue string) string {
...
 	if strings.HasPrefix(urlValue, forward) {
-		serviceName = urlValue[len(forward):]
+		// With a path like `/function/xyz/rest/of/path?q=a`, the service
+		// name we wish to locate is just the `xyz` portion.  With a postive
+		// match on the regex below, it will return a three-element slice.
+		// The item at index `0` is the same as `urlValue`, at `1`
+		// will be the service name we need, and at `2` the rest of the path.
+		matcher := functionMatcher.Copy()
+		matches := matcher.FindStringSubmatch(urlValue)
+		if 3 == len(matches) {
```
ivanayov: As in Alex's comment bellow, let's not use a number for this check, but edit for better readability. 
One may be confused why there's a check for length of 3 to take the 2nd element and may edit and break it in future.

telackey:
@ivanayov 

> will you have the chance to finish the PR or you'd rather delegate this?

If you'd like to finish it up with your notes, I have not objection at all.

gateway/handlers/forwarding_proxy.go
ivanayov: Addressed in #829

gateway/queue/types.go
alexellis: > I am not in a good position to test async functionality

What am I missing? Testing the async call is as simple as deploying OpenFaaS on a new VM, deploying your gateway and then invoking the function. Check the logs on the queue-worker and you're good.

gateway/server.go
```diff
@@ -170,7 +181,7 @@ func main() {
 
 	metricsHandler := metrics.PrometheusHandler()
 	r.Handle("/metrics", metricsHandler)
-	r.HandleFunc("/healthz", handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver)).Methods(http.MethodGet)
+	r.HandleFunc("/healthz", handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver, urlTransformer)).Methods(http.MethodGet)
```
alexellis: Why does `healthz` take a `urlTransformer`? Surely it's only needed for the function endpoints and if it is un-used, could we pass `nil`?

alexellis:
Sorry for the misunderstanding but we don't "delegate" finishing/fixings PRs to other contributors.

@telackey if you can rebase your PR ASAP I'll merge it and then sort out the other feedback.

In the meantime I've asked @ivanayov to test the `async-function` path-passing since you said you're not in a position to test it.

Alex

gateway/queue/types.go
telackey: Surely the question I was responding to about "equivalent change lined-up in nats-queue-worker" must also mean changing, building, and deploying the new nats-queue-worker too.

alexellis:
Please could you explain the difference between these two transformers?

```
	urlTransformer := handlers.FunctionPathTruncatingURLPathTransformer{}
	functionURLTransformer = handlers.FunctionPrefixTrimmingURLPathTransformer{}
```

I understand that the `functionURLTransformer` will trim the /function/ prefix from calls upstream, but why is a `urlTransformer` needed on the calls which route to `/system/functions` etc and to `healthz`?

telackey:
The path truncating one strips out the path as did the previous behavior.

telackey:
@alexellis @ivanayov 

Rebased.  Thank you both for all your help!

alexellis:
I've opened a new PR to finish the work.. please can you check everything is still as expected?

alexellis:
Moved over to #832

PR #816, opened 2018-08-14
Propagate Host header from gateway request to the function.

dmrub:
Host HTTP header was not propagated to the function because it is not
a part of http.Request.Header map.

## Description

This PR is described in issue https://github.com/openfaas-incubator/of-watchdog/issues/22

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/gateway/handlers/forwarding_proxy.go
+++ b/gateway/handlers/forwarding_proxy.go
@@ -53,6 +53,9 @@ func buildUpstreamRequest(r *http.Request, url string) *http.Request {
 	}
 
 	upstreamReq, _ := http.NewRequest(r.Method, url, nil)
+	if len(r.Host) > 0 {
+		upstreamReq.Host = r.Host
+	}
 	copyHeaders(upstreamReq.Header, &r.Header)
 
 	upstreamReq.Header["X-Forwarded-For"] = []string{r.RemoteAddr}
--- a/gateway/handlers/forwarding_proxy_test.go
+++ b/gateway/handlers/forwarding_proxy_test.go
@@ -68,6 +68,40 @@ func Test_buildUpstreamRequest_NoBody_GetMethod_NoQuery(t *testing.T) {
 
 }
 
+func Test_buildUpstreamRequest_HasHostHeaderWhenSet(t *testing.T) {
+	srcBytes := []byte("hello world")
+
+	reader := bytes.NewReader(srcBytes)
+	request, err := http.NewRequest(http.MethodPost, "http://gateway/function?code=1", reader)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	upstream := buildUpstreamRequest(request, "/")
+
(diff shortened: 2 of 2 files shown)


dmrub:
In Request structure for asynchronous processing Host field is missing, so is it safe to add it ? I am not familiar with json marshalling in Go (in particular I am Golang beginner) so are there any backwards compatibility problems possible ?

alexellis:
None that I expect. Just extend the struct/type with a string value called `Host` and then pick it up on the other side in the queue-worker project. Happy to give more guidance on Slack if you have joined?

gateway/handlers/queueproxy_test.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (q SingletonRequestQueue) GetRequest() (*queue.Request, error) {
+	if q.request == nil {
+		return nil, errors.New("SingletonRequestQueue is not initialized")
+	}
+	return *q.request, nil
+}
+
+func Test_MakeQueuedProxy(t *testing.T) {
```
alexellis: Can you explain what's happening in this test?

dmrub: In this test I want to test conversion of `http.Request` to `queue.Request` performed by HTTP handler function returned by `MakeQueuedProxy`. In order to do this I create first mock queue struct `SingletonRequestQueue`, which implements `queue.CanQueueRequests` interface in order to pass it to `MakeQueuedProxy`.
I only need to store one single `queue.Request` instance in my queue, which I get back for comparison via `GetRequest` func. Since `queue.CanQueueRequests` is passed to `MakeQueuedProxy` by value I can't just store a pointer to `queue.Request`. Instead I create another level of indirection and store pointer to pointer to `queue.Request`. This require initialization of `SingletonRequestQueue` by allocating a pointer to `queue.Request`: `SingletonRequestQueue{request: new(*queue.Request)}` plus checks that queue is initialized in all accessing funcs.
Because I want to check that function name is correctly stored to `queue.Request` I need to create and initialize gorilla/mux router, for this I used an approach based on this https://github.com/quii/testing-gorillas .

stefanprodan:
@alexellis @dmrub this change breaks the HTTP client https://github.com/openfaas/faas/issues/847 when running in a service mesh like Istio or if anyone uses a L7 proxy between the gateway and functions. The host is used by the reverse proxy to determine the destination. The Referer HTTP header should be used to forward the host value or something like `X-Origin-Host`

alexellis:
This PR was shared with the community for review with no objections.

I don't mind if we want to change this to use Referrer HTTP headers. @dmrub what are your thoughts?

stefanprodan:
Yes I've seen this PR before it got merged but it didn't click in my head until last night when I've starting working on the Istio integration.

dmrub:
I have no objections using different header for passing Host header to the function, but using Referer would hide Referer passed to gateway.

PR #843, opened 2018-09-02
Pass basic auth to all system calls

### LINKED ISSUE #836: Pass basic auth upstream to providers
## Expected Behaviour

When calling the /system/ or protected routes on the backend providers (upstream) i.e. faas-netes / faas-swarm, we should pass the basic credentials.

## Current Behaviour

This is not being done. For faas-swarm it is essential since faas-swarm is accessible via the network from functions. Related issue: https://github.com/openfaas/faas-swarm/issues/31

## Possible Solution

Add a wrapper for the protected routes which injects the basic auth headers upstream.

burtonr:
## Description

This changeset enables passing the basic auth credentials
to all /system/ calls to allow upstream providers to
perform authorization checks independent of the gateway.

This is essential for some providers, like Swarm, where
the system is accessible on the same network, and not
protected via the gateway

Signed-off-by: Burton Rheutan <rheutan7@gmail.com>

## Motivation and Context

Resolves #836 

This will enable the work on faas-swarm https://github.com/openfaas/faas-swarm/issues/31

This will allow providers to perform authorization checks when being called directly for additional security. 

## How Has This Been Tested?

I've run this locally running Swarm to verify that the additional headers are being passed, but without the other services using this header, it is a non-breaking change.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -89,6 +89,8 @@ func main() {
 	alertHandler := plugin.NewExternalServiceQuery(*config.FunctionsProviderURL)
 	faasHandlers.Alert = handlers.MakeAlertHandler(alertHandler)
 
+	infoHandler := handlers.MakeInfoHandler(handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver, nilURLTransformer))
+
 	if config.UseNATS() {
 		log.Println("Async enabled: Using NATS Streaming.")
 		natsQueue, queueErr := natsHandler.CreateNatsQueue(*config.NATSAddress, *config.NATSPort, natsHandler.DefaultNatsConfig{})
@@ -117,6 +119,8 @@ func main() {
 			handlers.DecorateWithBasicAuth(faasHandlers.ListFunctions, credentials)
 		faasHandlers.ScaleFunction =
 			handlers.DecorateWithBasicAuth(faasHandlers.ScaleFunction, credentials)
+		infoHandler = handlers.DecorateWithBasicAuth(infoHandler, credentials)
+		queryFunction = handlers.DecorateWithBasicAuth(queryFunction, credentials)
 	}
 
 	r := mux.NewRouter()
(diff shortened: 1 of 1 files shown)


gateway/server.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) Alex Ellis 2017. All rights reserved.
```
alexellis: Why have you removed my personal copyright from this file?

gateway/server.go
```diff
@@ -117,6 +119,12 @@ func main() {
 			handlers.DecorateWithBasicAuth(faasHandlers.ListFunctions, credentials)
 		faasHandlers.ScaleFunction =
 			handlers.DecorateWithBasicAuth(faasHandlers.ScaleFunction, credentials)
+		faasHandlers.Alert =
```
alexellis: If I merge this change then the auto-scaling will fail because Prometheus does not send auth strings to the scaling endpoint.

gateway/server.go
```diff
@@ -117,6 +119,12 @@ func main() {
...
+			handlers.DecorateWithBasicAuth(faasHandlers.Alert, credentials)
```
alexellis: I think this would break since I'm not sure the queue worker sends auth

gateway/server.go
```diff
@@ -117,6 +119,12 @@ func main() {
...
+		faasHandlers.AsyncReport =
+			handlers.DecorateWithBasicAuth(faasHandlers.AsyncReport, credentials)
+		infoHandler = handlers.DecorateWithBasicAuth(infoHandler, credentials)
+		queryFunction = handlers.DecorateWithBasicAuth(queryFunction, credentials)
```
alexellis: Did this not have auth enabled previously?

gateway/server.go
burtonr: I only changed it to the broader "OpenFaaS Project" to match other files as I thought there weren't supposed to be individual names in the copyright. I will add your name back, in addition to the overarching "OpenFaaS Author(s)" line

burtonr:
@alexellis I've updated to exclude the `Alert` and `AsyncReport` handlers from the auth check. Also, fixed the copyright line to retain your name (as was originally) and added "OpenFaaS Author(s)"

In doing so, I first verified that, yes those changes caused the scaling to no longer function. I've verified that the scaling _does_ function as expected now. 

The only 2 calls that are added to the auth check are the info and the query.

Issue #841, opened 2018-08-30
Sample-functions: Convert ADD https:// to use curl


alexellis:
## Expected Behaviour

All functions to use `curl` to get the watchdog since this can be cached by Docker.

```
RUN apk --no-cache add make curl \
    && curl -sL https://github.com/openfaas/faas/releases/download/0.9.0/fwatchdog > /usr/bin/fwatchdog \
    && chmod +x /usr/bin/fwatchdog
```

## Current Behaviour

Some use `ADD https://` which cannot be cached.

## Possible Solution

Update all samples in the faas repo and test that they still build.

rdimitrov:
I'll be happy to cover it.

alexellis:
Thank you. All yours.

rdimitrov:
@alexellis Do you have any suggestions about how to handle [this](https://github.com/rdimitrov/faas/blob/eceac0a097d4df0f428f1e724c84f23b990acb76/sample-functions/BaseFunctions/golang/Dockerfile.win#L9l) Windows Dockerfile?

rdimitrov:
* All samples for each architecture (where applicable) are updated and tested that there's no difference in terms of building. 
    * The only one left is [this one](https://github.com/rdimitrov/faas/blob/eceac0a097d4df0f428f1e724c84f23b990acb76/sample-functions/BaseFunctions/golang/Dockerfile.win#L9l)
* The changes related to each architecture are split in different commits - for ex. amd64, arm64, armhf and there's one additional for updating the docs.  

_Note: When the image is not Alpine, the RUN command is slightly changed due to the different package managers._

alexellis:
Hi @rdimitrov I don't think anyone is using the Windows example so I'd leave it as it is for now.

alexellis:
Do you have a pull request ready?

alexellis:
@rdimitrov in your next PR please add the text: "Closes #...."

PR #846, opened 2018-09-03
radoslav/sample functions - Use curl instead of ADD for getting the watchdog

rdimitrov:
## Description

The following PR will enable all functions to use curl to get the watchdog since this can be cached by Docker.
## Motivation and Context

The issue related to this PR is https://github.com/openfaas/faas/issues/841

## How Has This Been Tested?

The change implies only to the way watchdog is being get, so the verification was made by simply using docker build to build successfully each image for each architecture (where available). 

## Types of changes

## Checklist:

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -35,7 +35,7 @@ RUN go get github.com/microcosm-cc/bluemonday && \
 
 RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .
 
-ADD https://github.com/openfaas/faas/releases/download/0.8.0/fwatchdog /usr/bin
+ADD https://github.com/openfaas/faas/releases/download/0.9.0/fwatchdog /usr/bin
 RUN chmod +x /usr/bin/fwatchdog
 
 ENV fprocess="/go/src/app/app"
@@ -61,7 +61,7 @@ Update the Docker stack with this:
 ```
 FROM alpine:latest
 
-ADD https://github.com/openfaas/faas/releases/download/0.8.0/fwatchdog /usr/bin
+ADD https://github.com/openfaas/faas/releases/download/0.9.0/fwatchdog /usr/bin
 RUN chmod +x /usr/bin/fwatchdog
 
 ENV fprocess="wc"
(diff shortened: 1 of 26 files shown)


DEV.md
```diff
@@ -35,8 +35,10 @@ RUN go get github.com/microcosm-cc/bluemonday && \
 
 RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .
 
-ADD https://github.com/openfaas/faas/releases/download/0.8.0/fwatchdog /usr/bin
-RUN chmod +x /usr/bin/fwatchdog
+RUN apt-get update && apt-get install -y make curl \
```
alexellis: I think for the example it's OK to keep `ADD` for brevity and then show the `curl` method as an optimization underneath.

sample-functions/AlpineFunction/Dockerfile.arm64
```diff
@@ -1,8 +1,8 @@
 FROM arm64v8/alpine:3.7
```
alexellis: FYI we can remove the prefix now for `arm64v8` because Docker released an official multi-arch image under the tag `alpine:3.7`

sample-functions/AlpineFunction/Dockerfile.armhf
```diff
@@ -1,7 +1,8 @@
 FROM alpine:3.7
 
-ADD https://github.com/openfaas/faas/releases/download/0.9.0/fwatchdog-armhf /usr/bin/fwatchdog
-RUN chmod +x /usr/bin/fwatchdog
+RUN apk --no-cache add make curl \
```
alexellis: Why do we need `make`?

rdimitrov: My bad, it was just copy-pasted. I'll discard it from the files where it's not needed.

sample-functions/AlpineFunction/Dockerfile.arm64
rdimitrov: Nice, I'll update that as well.

DEV.md
rdimitrov: I agree, I'll update it then.

alexellis:
Merged. Thank you very much @rdimitrov 🎉

Issue #919, opened 2018-10-17
Feature: proxy HTTP redirect requests for synchronous requests


alexellis:
## Expected Behaviour

The API Gateway could proxy HTTP "redirect" requests for synchronous requests. These happen when a function returns a HTTP redirect status code plus a `Location: scheme://url` header.

The of-watchdog and stateless microservices can return custom / bespoke HTTP headers.

This does not apply or make any sense for asynchronous invocations.

## Current Behaviour

This technically "works as designed" in that any Location/redirect request is followed by the Golang proxying code and therefore if a function asks to redirect to https://google.com -> the API gateway's proxy code will download that page and return it.

## Possible Solution

I implemented this for the OpenFaaS Cloud Auth service, the same approach can be applied to the http.Client. https://github.com/openfaas/openfaas-cloud/blob/3d3ed09707817b436751b279973d6f26191930a4/router/main.go#L34

## Steps to Reproduce (for bugs)

1. Use of-watchdog or a stateless microservice
2. Send a HTTP header of "Location:" and a HTTP status code to redirect
3. See that page returned but on the location / address of the function.

I.e. with [node8-express](https://github.com/openfaas-incubator/node8-express-template):

```js
"use strict"

module.exports = (event, context) => {
  context
    .headers({'Location': 'https://www.google.com/'})
    .status(307)    // Temporary
    .succeed('Page has moved.')
}
```

## Context

Reported by a user who wanted to do OAuth redirects from a function. https://github.com/openfaas-incubator/of-watchdog/issues/34

The watchdog / of-watchdog may need an additional change, but the gateway can be proven out just with a container exposing port 8080. https://docs.openfaas.com/reference/workloads/

It goes without saying that this needs to be fully tested and with not just the "happy path" scenario.

alexellis:
I've tried this with a stateless microservice adapted from the node8-express template, and the code patch above. It worked :+1: 

Separate testing/changes may be needed for the watchdogs.

alexellis:
Derek close: released.

PR #925, opened 2018-10-19
Don't follow redirects from functions

alexellis:
- Covers part of 919 by making the HTTP client used for proxying
stop following redirects. Tested with a stateless microservice,
but additional code changes may be requierd in the queue-worker,
the watchdogs and other areas.

Tested on Swarm with stateless microservice (Node.js) issuing
a redirect via Location header.


alexellis:
Ref: https://github.com/openfaas/faas/issues/919

PR #929, opened 2018-10-20
Create user/pass secrets and add help/noAuth flags to deploy_stack.ps1

### LINKED ISSUE #928: Windows deploy_stack.ps1 does not setup basic auth secrets
## Expected Behaviour

Windows `deploy_stack.ps1` should set the `basic-auth-user` and `basic-auth-password` secrets if absent.

## Current Behaviour

The `deploy_stack.ps1` ignores the secrets.

## Possible Solution

Update the Powershell script to prompt for the user/password if not already created.

1. Run `deploy_stack.ps1` on a Windows Docker Swarm with/without the http basic auth secrets created.

## Context

Currently using Windows 10.

## Your Environment

- using the script from the latest on master (at the time of raising the issue openfaas/faas@0a5e343c5de4acc697d89542a79d204edaf39c9)

johnmccabe:
## Description

This PR updates the Windows `deploy_stack.ps1` Powershell script to set the `basic-auth-user` and generate the `basic-auth-password` secrets if absent.

It also adds the following flags:

- `-help`, `-h`: print help text
- `-noAuth`, `-n`: disable basic auth

## Motivation and Context

- [x] I have raised an issue to propose this change (Closes #928)

## How Has This Been Tested?

Deployed on Windows 10 with the following scenarios:
- No user or pass secret
- User secret exists, pass missing
- Pass secret exists, user missing
- Both user and pass secrets exist

## Types of changes

## Checklist:

### DIFF
--- a/deploy_stack.ps1
+++ b/deploy_stack.ps1
@@ -1,44 +1,66 @@
 #!ps1
 
+param (
+    [switch] $noAuth,
+    [switch] $n
+)
+
 if (Get-Command docker -errorAction SilentlyContinue)
 {
     docker node ls 2>&1 | out-null
+    if(-Not $?)
+    {
+        throw "Docker not in swarm mode, please initialise the cluster (`docker swarm init`) and retry"
+    }
+
+    Add-Type -AssemblyName System.Web
+    $secret = [System.Web.Security.Membership]::GeneratePassword(24,5)
+    $user = 'admin'
+
+    Write-Host "Attempting to create credentials for gateway.."
+    $user_secret = "basic-auth-user"
+    docker secret inspect $user_secret 2>&1 | out-null
     if($?)
     {
-        $user_secret = "basic-auth-user"
-        docker secret inspect $user_secret 2>&1 | out-null
-        if($?)
-        {
-            Write-Host "$user_secret secret exists"
-        }
-        else
-        {
-            $user = Read-Host 'Admin User?'
-            $user | docker secret create $user_secret -
-        }
-
-        $password_secret = "basic-auth-password"
-        docker secret inspect $password_secret 2>&1 | out-null
-        if($?)
-        {
-            Write-Host "$password_secret secret exists"
-        }
-        else
-        {
-            $pass = Read-Host 'Password?' -AsSecureString
-            [Runtime.InteropServices.Marshal]::PtrToStringAuto([Runtime.InteropServices.Marshal]::SecureStringToBSTR($pass)) | docker secret create $password_secret -
-        }
-
-        Write-Host "Deploying stack"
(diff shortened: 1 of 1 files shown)


johnmccabe:
This Closes #928

alexellis:
Is it possible to match the flags of the bash script?

deploy_stack.ps1
```diff
@@ -1,11 +1,76 @@
...
 {
-    Write-Host "Deploying stack"
+    docker node ls 2>&1 | out-null
+    if(-Not $?)
+    {
+        throw "Docker not in swarm mode, please initialise the cluster (`docker swarm init`) and retry"
+    }
+
+    Add-Type -AssemblyName System.Web
+    $secret = [System.Web.Security.Membership]::GeneratePassword(24,5)
```
alexellis: For consistency, please can you hash this value in the same way as the Linux bash script so that we don't have two sets of behaviour?

johnmccabe:
Can you clarify what you mean by hash here, are you suggesting making the password use hex only?

alexellis:
I.e. mirror the behaviour of the bash script.

It's something like this (paraphrased):

```
$ head /dev/urandom | shasum
935ac4308a8691851faa8b45ae787a7585550b30  -
```

alexellis:
Is `--no-auth` also doable with powershell so that we just have one flag for both?

deploy_stack.ps1
johnmccabe: Whats the rational for making the generated password *look* like the Linux equivalent (which is due to the method chosen for generating) - the `GeneratePassword` command is the PowerShell way of generating a valid password.

It *should* be possible to hash the password, but it complicates the script and I don't think its desirable.

alexellis:
The cluster components are native Linux, all we're changing is the client used to introduce the secret. I.e. git bash or WSL users may use the bash script.

How about? https://gist.github.com/benrobot/67bacea1b1bbe4eb0d9529ba2c65b2a6

johnmccabe:
I don't follow your thinking here.

alexellis:
I'm not a powershell expert so I'll l defer to your judgement on the flag style. I would like to see the password hashed so that we have consistent values between Linux, Mac and Windows whether on swarm or Kubernetes. if we have general concerns about the length of the hash (which is widely in use) and then please open an issue about that so we can address it for all the users. 

Would that work for you in terms of moving this PR forward?

alexellis:
Out of the approaches I've seen to generate a hash this looks to be the most concise:

https://gist.github.com/WalternativE/450b155c45f81b14290f8ded8324a283

Plus `.toLower()` at the end to match the format we get from bash on Linux/Mac/Windows.

PR #949, opened 2018-11-05
Enable backoff/retries on scaling up

alexellis:
Enable backoff/retries on scaling up

## Description

This change is needed for Docker Swarm which may give an error
when several concurrent requests come in to scale a deployment.

Docker will throw an error about being out of sync which could
have prevented some requests from being served.

## Motivation and Context

No issue is being raised, but this was found whilst working on
https://github.com/openfaas/nats-queue-worker/issues/32 with
high concurrency testing and watching the logs / completed
request count.

## How Has This Been Tested?

Tested on Docker Swarm before/after with the hey tool and figlet
scaled down to zero replicas.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/scaling/function_scaler.go
+++ b/gateway/scaling/function_scaler.go
@@ -57,12 +57,31 @@ func (f *FunctionScaler) Scale(functionName string) FunctionScaleResult {
 			minReplicas = queryResponse.MinReplicas
 		}
 
-		log.Printf("[Scale] function=%s 0 => %d requested", functionName, minReplicas)
+		scaleResult := backoff(func(attempt int) error {
+			queryResponse, err := f.Config.ServiceQuery.GetReplicas(functionName)
+			if err != nil {
+				return err
+			}
 
-		setScaleErr := f.Config.ServiceQuery.SetReplicas(functionName, minReplicas)
-		if setScaleErr != nil {
+			f.Cache.Set(functionName, queryResponse)
+
+			if queryResponse.Replicas > 0 {
+				return nil
+			}
+
+			log.Printf("[Scale %d] function=%s 0 => %d requested", attempt, functionName, minReplicas)
+			setScaleErr := f.Config.ServiceQuery.SetReplicas(functionName, minReplicas)
+			if setScaleErr != nil {
+				return fmt.Errorf("unable to scale function [%s], err: %s", functionName, setScaleErr)
+			}
+
+			return nil
+
+		}, int(f.Config.SetScaleRetries), f.Config.FunctionPollInterval)
+
+		if scaleResult != nil {
 			return FunctionScaleResult{
-				Error:     fmt.Errorf("unable to scale function [%s], err: %s", functionName, err),
+				Error:     scaleResult,
 				Available: false,
 				Found:     true,
 				Duration:  time.Since(start),
@@ -105,4 +124,24 @@ func (f *FunctionScaler) Scale(functionName string) FunctionScaleResult {
 		Found:     true,
 		Duration:  time.Since(start),
 	}
+}
+
+type routine func(attempt int) error
+
+func backoff(r routine, attempts int, interval time.Duration) error {
+	var err error
+
+	for i := 0; i < attempts; i++ {
+		res := r(i)
+		if res != nil {
+			err = res
+
+			log.Printf("Attempt: %d, had error: %s\n", i, res)
+		} else {
+			err = nil
+			break
+		}
+		time.Sleep(interval)
+	}
+	return err
 }
--- a/gateway/scaling/function_scaler.go
+++ b/gateway/scaling/function_scaler.go
@@ -38,6 +38,17 @@ type FunctionScaleResult struct {
 // the minimum replicas metadata
 func (f *FunctionScaler) Scale(functionName string) FunctionScaleResult {
 	start := time.Now()
+
+	if cachedResponse, hit := f.Cache.Get(functionName); hit &&
+		cachedResponse.AvailableReplicas > 0 {
+		return FunctionScaleResult{
+			Error:     nil,
+			Available: true,
+			Found:     true,
+			Duration:  time.Since(start),
+		}
+	}
+
 	queryResponse, err := f.Config.ServiceQuery.GetReplicas(functionName)
 
 	if err != nil {
@@ -90,7 +101,9 @@ func (f *FunctionScaler) Scale(functionName string) FunctionScaleResult {
 
 		for i := 0; i < int(f.Config.MaxPollCount); i++ {
 			queryResponse, err := f.Config.ServiceQuery.GetReplicas(functionName)
-			f.Cache.Set(functionName, queryResponse)
+			if err == nil {
+				f.Cache.Set(functionName, queryResponse)
+			}
 			totalTime := time.Since(start)
 
 			if err != nil {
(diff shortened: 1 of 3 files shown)


alexellis:
For testing a docker image exists `openfaas/gateway:0.9.9-rc2`

```
docker service update func_gateway --image=openfaas/gateway:0.9.9-rc2

faas store deploy figlet
docker service scale figlet=0

hey -n 100 -q 10 -c 10 http://127.0.0.1:8080/function/figlet --data test -m POST
```

You should see less than x100 HTTP 200 requests if done with the older version along with some logs in func_faas-swarm and x100 HTTP 200 with the newer version.

alexellis:
I need to add back the use of the cache before attempting to query the upstream provider `f.Cache.Get` which is seen here -> https://github.com/openfaas/faas/commit/9cea08c7288b88fdda2bb2c5bbfc3e42130250ce#diff-bfc80cebb3aae36705f2d21c0a5c2302L43

gateway/server.go
```diff
@@ -137,7 +137,8 @@ func main() {
 	if config.ScaleFromZero {
 		scalingConfig := scaling.ScalingConfig{
 			MaxPollCount:         uint(1000),
-			FunctionPollInterval: time.Millisecond * 10,
+			SetScaleRetries:      uint(20),
```
LucasRoesler: `SetScaleRetries` sounds like it should be a function, perhaps this should just be `MaxScaleRetries`

gateway/scaling/function_scaler.go
```diff
@@ -106,3 +139,23 @@ func (f *FunctionScaler) Scale(functionName string) FunctionScaleResult {
 		Duration:  time.Since(start),
 	}
 }
+
+type routine func(attempt int) error
+
+func backoff(r routine, attempts int, interval time.Duration) error {
```
LucasRoesler: we should really consider adding `ctx` to this and checking that the `ctx` is cancelled/done and also passing it to the routine so that it can also be used to cancel work or to support tracing. Context is generally most useful for methods that talk to external services, which is a big part of why `backoff` exists. It is a natural place to put `ctx` and even if we are not passing it around everywhere, it is clearly a place it would be wanted in the future.

gateway/scaling/function_scaler.go
```diff
@@ -106,3 +139,23 @@ func (f *FunctionScaler) Scale(functionName string) FunctionScaleResult {
 		Duration:  time.Since(start),
 	}
 }
+
+type routine func(attempt int) error
```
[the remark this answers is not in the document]
alexellis: This is here for logging. I'll have a think whether it could be moved. Logging is not required for every routine a `backoff()` function can run. So this moves the responsibility (as you mentioned) to the routine to decide.

gateway/server.go
alexellis: `MaxScale` seems to have some ambiguity with the maximum replicas/scale allowed for a function. `SetScale` is referring to the upstream call. Do you think there's another name that can fit here?

gateway/scaling/function_scaler.go
```diff
@@ -57,12 +68,31 @@ func (f *FunctionScaler) Scale(functionName string) FunctionScaleResult {
...
+				return err
+			}
+
+			f.Cache.Set(functionName, queryResponse)
+
+			if queryResponse.Replicas > 0 {
+				return nil
+			}
+
+			log.Printf("[Scale %d] function=%s 0 => %d requested", attempt, functionName, minReplicas)
```
alexellis: @LucasRoesler this is the other issue with not allowing the routine to know about the attempt - since here we can access variables within the closure that we couldn't in an other loop/function.

gateway/scaling/function_scaler.go
```diff
@@ -38,6 +38,17 @@ type FunctionScaleResult struct {
 // the minimum replicas metadata
 func (f *FunctionScaler) Scale(functionName string) FunctionScaleResult {
 	start := time.Now()
+
+	if cachedResponse, hit := f.Cache.Get(functionName); hit &&
```
LucasRoesler: This means that a scale can only occur as often as `CacheExpiry` ?

gateway/server.go
LucasRoesler: `MaxScaleAttempts` maybe

alexellis: MaxScale also has the same connotation as before, max scale is a label we set for controlling scaling. If `ScaleUpAttempts` seems better to you I could change to that?

gateway/scaling/function_scaler.go
alexellis: Whilst `Context` objects are used for this often in Golang code, the context is not consumable by any of the APIs we're calling. It's for that reason that I can't see how this would help right now. Perhaps you would be interested in doing some refactoring on the APIs to enable it to be consumed after this PR is merged?

My intent with `backoff()` is to resolve the current issue, not to create a public package for general/generic use, but if this pattern is useful I think we could refactor it and extract it later.

gateway/scaling/function_scaler.go
alexellis: If you find where `Scale` is called from (and when) you should see that it encapsulates the whole workflow including whether Scaling is needed. This is instead of two methods for: `IsScaled()` (i.e. should we scale) and `Scale()` (let's scale) being split out into two separate methods.

alexellis: To put it another way - if a cache is not used, it means we'd be calling the provider's function status method for every single invocation of a function which adds considerable latency and resource consumption.

gateway/server.go
LucasRoesler: I can't think of anything that is clearly better so we might as well leave it alone

gateway/scaling/function_scaler.go
LucasRoesler: ok

gateway/scaling/function_scaler.go
LucasRoesler: ok

gateway/scaling/function_scaler.go
alexellis: I don't want it to sound like I'm not listening to your idea. I spent a couple of hours +/- trying to retrofit Context into the current usage. It didn't seem to make sense with the underlying APIs.

gateway/scaling/function_scaler.go
alexellis: Given you didn't want to see the attempt integer in the routine, how would you approach this problem?

alexellis:
@LucasRoesler we need to release this version to add the caching on available replicas back in which was inadvertently removed in the previous minor release. Given your feedback I'll look at whether `Scale` / `IsAvailable` can or need to be split out before being consumed in the queue-worker.

gateway/scaling/function_scaler.go
LucasRoesler: I understand

gateway/scaling/function_scaler.go
LucasRoesler: I would have passed a value to the backoff directly

```golang
func backoff(name string, r routine, attempts int, interval time.Duration) error {
	var err error
 	for i := 0; i < attempts; i++ {
		res := r(i)
		if res != nil {
			err = res
 			log.Printf("%s attempt: %d of %d, had error: %s\n", name, attempts, i, res)
		} else {
 			log.Printf("%s attempt: %d of %d, succeeded\n", name, attempts, i)
			err = nil
			break
		}
		time.Sleep(interval)
	}
	return err
}
```

and then, I would remove the log statement from the routine completely

```golang
routineName := fmt.Sprintf("[Scale] function=%s 0 => %d", functionName, minReplicas)
scaleResult := backoff(routineName, func() error {
...
 }
```

Issue #807, opened 2018-08-07
Proposal: manage secrets via OpenFaaS API


alexellis:
## Feature

This feature would allow secrets to be created in the backing provider's credential store using a single CLI command regardless of what the backing orchestrator is.

```
faas-cli secret create --name secret-name --value secret-value
```

It could also be used with reading from a file with `--from-file` for instance.

## Why is this needed?

Each tutorial for OpenFaaS has to list Docker Swarm and Kubernetes secret creation commands. This is becoming hard to manage and duplicates effort. It'd be better if we could abstract this away.

A secrets API would also be useful for automating OpenFaaS.

DELETE / POST / GET / PUT -> `/system/secrets`

I would suggest that fetching a secret value would be not allowed, but getting a list of secrets, deleting or updating should be.

Edit: `GET` is the same as a `list` operation, it is not meant to get the plain-text value of a secret, that is not what this feature is about (this was misinterpreted in the comments)

## Possible Solution

Add support in OpenFaaS:

* Create REST API extensions
* Update Swarm, K8s providers
* Update CLI

## Steps to Reproduce (for bugs)

Example: https://docs.openfaas.com/reference/secrets/#use-the-secret-in-your-function

LucasRoesler:
I think this is a good idea, I like that it will streamline the documentation. I don't see any major gaps at the moment.  

There might be some questions around access controls, but the OpenFaaS system wouldn't be able to support those right now anyway. The system would not be any more leaky than it currently is.

rgee0:
I like the idea, and like Lucas my first thought was around controls.  There would be a shift here in managing secrets which would need some thought, particularly in the context of dev teams sharing a platform.  It feels like this is something which could do with multi-tenancy support in order to work effectively.

alexellis:
I think I'm hearing: exposing a secrets creation endpoint feels like opening up more access, but if you have access to the API you can do whatever you want already. TLS + auth should be a must for using this - it's one of the reasons I think we should have list, but not "get value"

I'd also like @stefanprodan's input.

stefanprodan:
If we want to offer secrets management with OpenFaaS I would consider the following:

Swarm doesn't support secret updates so we need to generate a unique name for it, detect what functions are using the secret, create a new function definition with the new secret name, change the mount, delete the current Docker service and create a new one. I see no way to avoid downtime for this, also the code will break since the secret name changes so does the path, an env var is needed for the function code to find the secret on disk.

[...]

PR #968, opened 2018-11-19
[WIP] - Feature/secret management to facilitate proposal #807

acornies:
## Description
This is the first round of changes to facilitate a simpler way of managing secrets by adding a new route and secret handler to be implemented in each respective provider.

## Motivation and Context
From proposal #807 

## How Has This Been Tested?
Not fully tested yet, only manual integration tests done in Vagrant/Docker Swarm environment. TODO: add integration tests in pre-defined location

AFAICT: This feature introduces a new handler for the providers to implement CLUD on secrets: faas-netes,swarm,nomad, faas-provider? faas-cli functionality also needs to be implemented (another PR outside of this scope)
## Types of changes

## Checklist:
- [ ] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ certifier
 .editorconfig
 /contrib/staging/
 **/*.sha256
+.vagrant
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,78 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# All Vagrant configuration is done below. The "2" in Vagrant.configure
+# configures the configuration version (we support older styles for
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure("2") do |config|
+  # The most common configuration options are documented and commented below.
+  # For a complete reference, please see the online documentation at
+  # https://docs.vagrantup.com.
+
+  # Every Vagrant development environment requires a box. You can search for
+  # boxes at https://vagrantcloud.com/search.
+  config.vm.box = "generic/ubuntu1604"
+
+  # Disable automatic box update checking. If you disable this, then
+  # boxes will only be checked for updates when the user runs
+  # `vagrant box outdated`. This is not recommended.
+  # config.vm.box_check_update = false
+
+  # Create a forwarded port mapping which allows access to a specific port
+  # within the machine from a port on the host machine. In the example below,
+  # accessing "localhost:8080" will access port 80 on the guest machine.
+  # NOTE: This will enable public access to the opened port
+  # config.vm.network "forwarded_port", guest: 80, host: 8080
+
+  # Create a forwarded port mapping which allows access to a specific port
+  # within the machine from a port on the host machine and only allow access
+  # via 127.0.0.1 to disable public access
+  config.vm.network "forwarded_port", guest: 8080, host: 8080, host_ip: "127.0.0.1"
+
+  # Create a private network, which allows host-only access to the machine
+  # using a specific IP.
+  # config.vm.network "private_network", ip: "192.168.33.10"
+
+  # Create a public network, which generally matched to bridged network.
+  # Bridged networks make the machine appear as another physical device on
+  # your network.
+  # config.vm.network "public_network"
+
+  # Share an additional folder to the guest VM. The first argument is
+  # the path on the host to the actual folder. The second argument is
+  # the path on the guest to mount the folder. And the optional third
+  # argument is a set of non-required options.
+  config.vm.synced_folder "./", "/vagrant"
+
+  # Provider-specific configuration so you can fine-tune various
+  # backing providers for Vagrant. These expose provider-specific options.
+  # Example for VirtualBox:
+  #
+  config.vm.provider "virtualbox" do |vb|
+    # Display the VirtualBox GUI when booting the machine
+    vb.gui = false
+  
+    # Customize the amount of memory on the VM:
+    vb.memory = "2048"
+  end
+
+  config.vm.provider "vmware_fusion" do |vmf|
+    vmf.gui = false
+    vmf.memory = "2048"
+  end
+  #
+  # View the documentation for the provider you are using for more
+  # information on available options.
+
+  # Enable provisioning with a shell script. Additional provisioners such as
+  # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
+  # documentation for more information about their specific syntax and use.
+  config.vm.provision "docker"
+
+  config.vm.provision "shell", inline: <<-SHELL
+    docker swarm init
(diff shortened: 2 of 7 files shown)


Vagrantfile
```diff
@@ -0,0 +1,78 @@
+# -*- mode: ruby -*-
```
alexellis: Any chance the Vagrantfile can sit in the contrib folder? 🙂

gateway/requests/requests.go
```diff
@@ -90,3 +90,8 @@ type AsyncReport struct {
 type DeleteFunctionRequest struct {
 	FunctionName string `json:"functionName"`
 }
+
+// SecretRequest create, update a secret
+type SecretRequest struct {
```
alexellis: How will the map work? CLUD using HTTP verbs?

acornies: Right - It should be along the lines of secretName secretValue, then L part of CLUD will return a list of this type minus the value

Vagrantfile
acornies: sure thing

gateway/requests/requests.go
alexellis: @ewilde what are your thoughts? It I have the impress that it may be slightly orthogonal to the existing design.

acornies: I've made some commits since the start of this thread. I reverted my last one since I think we want to stick with one secret/request (POST,PUT,DELETE) approach.

gateway/requests/requests.go
```diff
@@ -90,3 +90,20 @@ type AsyncReport struct {
...
+
+// DeleteSecretRequest remote a secret by name
+type DeleteSecretRequest struct {
+	SecretName string `json:"secretName"`
+}
+
+// Secret schema use Value only in write-only http verbs
+type Secret struct {
+	Name        string             `json:"name"`
+	Value       string             `json:"value"` // write-only
```
alexellis: @stefanprodan @LucasRoesler should this be `string` or `[]byte`?

gateway/requests/requests.go
```diff
@@ -90,3 +90,20 @@ type AsyncReport struct {
...
+	Annotations *map[string]string `json:"annotations"`
```
alexellis: I think annotations or metadata will be useful and necessary. Do both Swarm and Kubernetes support adding this to secrets?

gateway/requests/requests.go
```diff
@@ -90,3 +90,20 @@ type AsyncReport struct {
 type DeleteFunctionRequest struct {
 	FunctionName string `json:"functionName"`
 }
+
+// CreateSecretRequest create a secret w/ annotations
+type CreateSecretRequest struct {
```
alexellis: Do we need a ListSecret equivalent or will that just be `[]Secret`?

gateway/requests/requests.go
acornies: byte array makes sense to me as that opens the door save json or other structure based on that key. Vault would support it anyway.

gateway/requests/requests.go
acornies: Looking at this again, perhaps we should move it to the CreateSecretRequest struct instead?

gateway/requests/requests.go
acornies: I was just thinking []Secret as the response type for GET.

gateway/requests/requests.go
alexellis: I am not sure if I'm right, just wanted to open discussion.

LucasRoesler: I agree, `[]byte` seems like a good choice

gateway/requests/requests.go
LucasRoesler: Why only on the Create? Won't we want to be able to read and display those annotations? 

Is there a reason to use a pointer here? `maps` behave like pointers already, e.g. 
```golang
var m map[string]string
m = nil
fmt.Printf("%d", len(m))
```
is valid and I think when you json unmarshal and the value is empty or omitted, the map will be set nil as well.

alexellis: It might be due to copy/paste from other areas of code? We should add omitempty on some of these properties too.

gateway/requests/requests.go
stefanprodan: I would go with a base64 encoded string, like K8s does.

LucasRoesler: That also makes sense. But we should make sure to include a good doc string here to make sure people remember that when they are using/implementing the secrets providers

LucasRoesler:
This should have a update to the swagger spec in the `api-doc` folder, right?  

One thing that I thought about while reviewing this is that the List endpoint might want to support filtering.  There is no way to specify this in the go code here because those values would live in the GET parameters.  But if we are keeping the swagger spec up-to-date, that spec should include a definition of the filter parameters that should be supported by the provider.

gateway/requests/requests.go
acornies: How about simply a `SecretRequest` type? Members could be:
```
Secret Secret `json:"secret"`
Annotations map[string]string `json:"annotations"`
```
POST, PUT are pretty straightforward. GET would then return `[]SecretRequest` instead, omitting certain values ;).

LucasRoesler: I would highly recommend separating the Request struct from the Response struct.  It becomes much easier to document and to extend in the future.

acornies:
My next pass should include stuff discussed here, plus some outstanding things on the TODO list. Standby...

gateway/requests/requests.go
alexellis: We should be able to represent whatever we need in a string here. Since the users won't see the value from the API we shouldn't need base64.

LucasRoesler:
@acornies I saw the latest push, please let me know when you feel it is ready for another review

acornies:
@LucasRoesler thanks - I still need to add integration tests, and do more testing with the faas-swarm provider to start.

My style is to push smaller incremental updates. I can hold off the PR until it's more complete next time.

LucasRoesler:
Hi @acornies , how are things going with the secrets API? If there is anything that can help with or if you have any questions I can help answer, please let me know.

acornies:
Thanks @lucasalexander will do. I had no time in November, early Dec. However, Now that things have died down a bit, I expect to make more progress. Might need help with k8s provider, but we'll see how it goes.

.gitignore
```diff
@@ -8,3 +8,4 @@ certifier
 .editorconfig
 /contrib/staging/
 **/*.sha256
+.vagrant
```
alexellis: This should probably be in the developer's .gitignore?

PR #976, opened 2018-11-22
Fix scale from zero to be true by default

ivanayov:
It should be configured to true by default and disabled by choice.


alexellis:
SGTM

PR #988, opened 2018-12-01
Download vendor cdn files for gateway

### LINKED ISSUE #982: UI: Bundle all required JS dependencies
## Expected Behaviour
The UI should work, in any environment, on any network.

## Current Behaviour
The UI relies on importing JS dependencies from the internet through `<script>` tags in the `index.html`, which means it does not work anywhere where there is no internet connection.

## Possible Solution
Bundling minified versions of the dependencies and loading them locally providing a self contained application capable of working within the ethos of Docker.

## Context
We are trying to run this in an environment with no internet connection.

Happy to make the simple changes myself and submit a PR.

burtonr:
## Description

This change set downloads the CDN resources for the gateway
and bundles them with the other static resources for the UI.

This is needed for situations where a user does not have access
to the CDN either because of firewall rules or network policy.

The files and versions remain the same, only now loaded locally
with the exception of the angular-material CSS which was updated
to match the other angular-material version.

Signed-off-by: Burton Rheutan <rheutan7@gmail.com>

## Motivation and Context

Resolves #982 

## How Has This Been Tested?

Built the gateway container locally and attached it to a running Swarm OpenFaaS cluster.
Open the network diagnostics to view the resources and ensure that no external calls are made (with the exception of the Google fonts)

screenshot from 2018-12-01 12-47-45

## Types of changes

## Checklist:


burtonr:
@alexellis @kenfdev Please have a look and let me know if there's anything else that would need to be done.

kenfdev:
Thank you @burtonr for taking your time to make this change!

> with the exception of the angular-material CSS which was updated to match the other angular-material version.

I agree with this change but not sure if we should do it at this PR. IMHO we should confirm that the scripts are completely the same as it was before.

Also, I wonder why there is some indents in some scripts... Could you tell me how you downloaded this?

Below is a capture of FileMerge for `angular-material.min.js` (left: before, right: this PR)

image

Again, thank you for taking this task and investing your time.

omrishtam:
You have a duplicate line of the script tag for the angular.min.js script at line 198-199 in gateway/assets/index.html

burtonr:
@omrishtam Good eye! Not sure how that happened, but I've removed that duplicate line

@kenfdev The process I took was opening the CDN links in a browser window and copying the text into a new file.I use Firefox as the browser, and use Ubuntu as the OS. Not sure where the indent came from. 

Thank you both for looking at it though!

omrishtam:
> Hey @omrishtam thanks for helping out. Do we have you on Slack yet.. would you like to join? Alex

Sure!

alexellis:
Join here -> https://docs.openfaas.com/community/

gateway/assets/index.html
```diff
@@ -3,7 +3,7 @@
 <head>
     <title>OpenFaaS Portal</title>
     <link href="https://fonts.googleapis.com/css?family=Rationale|Roboto+Mono|Roboto:300,400,400i,500,700" rel="stylesheet">
```
alexellis: @kenfdev can we do anything for this? What would happen in a fully offline mode?

LucasRoesler:
Can we add the `*min.js.map` files too?

LucasRoesler:
Using `openssl md5 **/*` I generated these signatures that we can verify against he cdn values

```
MD5(angularjs/1.6.9/angular-animate.min.js)= 221238cdc62fb23ff08a73c1bf939b24
MD5(angularjs/1.6.9/angular-aria.min.js)= cce957e52a4d0c9514b4e338e4aa82fb
MD5(angularjs/1.6.9/angular-messages.min.js)= 59a5220074801e00acf4bc3a3d42fa6a
MD5(angularjs/1.6.9/angular.min.js)= ca6464a8e26218520f4b0c3bf2137013
MD5(angular_material/1.1.4/angular-material.min.js)= abc1463540993dda16fe7f81a3caddfc
```

LucasRoesler:
it might be nice to have a little bash script included in this PR to make it easy to redownload the scripts and also generate these hashes in the future

kenfdev:
> Can we add the *min.js.map files too?

Curious why we need the `*min.js.map` files Lucas. IIUC, these are for sourcemaps to unfold the `min.js` scripts. So I don't think we need these. Am I missing something?

> Only the angular.min.js file matches. What are the differences?

https://github.com/openfaas/faas/pull/988#issuecomment-443484151
^It's probably related to the indents.

> it might be nice to have a little bash script included in this PR to make it easy to redownload the scripts and also generate these hashes in the future

I totally agree with this!

burtonr:
@LucasRoesler and @kenfdev There must have been some sort of magical formatting happening when I used `ctrl+C` and `ctrl+V` to get the file content originally causing the extra indent, which invalidated the md5 hash.

I've updated the files using the `curl` commands Lucas used and verified the md5 hashes now match. 

I did not add a script to automatically add these files intentionally as it is an extra piece not necessarily needed to make the gateway work offline. I definitely agree we should add something to do that, but I think it should be a separate PR so that conversation and changes won't delay the original issue.

LucasRoesler:
@kenfdev I think adding the source maps improves the developer experience because errors then point to the unminified code in the console.  They aren't needed, but definitely nice when an error starts from inside the vendored minified libraries

LucasRoesler:
I think that this looks good now.  

I have no issue with not including a script to streamline this, but I will insist on it when we decide to update the libraries again

kenfdev:
>  I think adding the source maps improves the developer experience because errors then point to the unminified code in the console. They aren't needed, but definitely nice when an error starts from inside the vendored minified libraries

Having the fact that all the scripts are Open Source and that source maps are only downloaded when one opens the dev tools, I agree with you (the only downside I can think of is that it adds about 500kb to the image).

That said, I guess we can conclude that it is not directly related to this issue for now. Thank you for clearing it out though 👍

alexellis:
I'd like to see the script too. Lucas maybe you could run with that?

PR #996, opened 2018-12-08
Fix fnc image name in the ui

### LINKED ISSUE #990: UI: Refresh image tag on periodic basis
## Expected Behaviour

The image tag should change when deploying a new version in the UI.

## Current Behaviour

The state has changed in the cluster but is cached on the UI for efficiency and shows the older tag version. This was confusing for one of our users "zeeque" who reported this on Slack.

## Possible Solution

We just need to bust the cache or refresh the objects when we refresh the list of functions.

1. Deploy a function
2. Update the tag
3. Deploy it again

screenshot 2018-12-04 at 09 58 09

LucasRoesler:
## Description

- Update the function image value during the `refreshData`
- Apply vscode default JS/TS formatting

### Note 
Please start your review with the commit 3fe8267 this contains the actual implementation change
The second commit is purely formatting changes.

## Motivation and Context

Resolves #990 

## How Has This Been Tested?

This manually tested in a local deployment
0. update docker-compose to mount the local assets folder,  adding this to the `gateway` definition
```yaml
        volumes:
            - "./gateway/assets:/home/app/assets"
```
1. deploy to docker swarm using the `./deploy_stack.sh`
2. deploy a project using `faas-cli up -f stack.yml --tag=sha`
3. open the function details in the ui
4. modify the function, commit, and redeploy using `faas-cli up -f stack.yml --tag=sha`
5. observe that the Image field updates to match the new git sha

## Types of changes

## Checklist:

### DIFF
--- a/gateway/assets/script/bootstrap.js
+++ b/gateway/assets/script/bootstrap.js
@@ -185,27 +185,25 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$interval', '$f
 
             var cl = function (previousItems) {
                 $http.get("../system/functions").then(function (response) {
-                    if (response == null || response.data == null) {
-                        return
-                    }
+                    if (response && response.data) {
+                        if (previousItems.length != response.data.length) {
+                            $scope.functions = response.data;
 
-                    if (previousItems.length != response.data.length) {
-                        $scope.functions = response.data;
-
-                        // update the selected function object because the newly fetched object from the API becomes a different object
-                        var filteredSelectedFunction = $filter('filter')($scope.functions, { name: $scope.selectedFunction.name }, true);
-                        if (filteredSelectedFunction && filteredSelectedFunction.length > 0) {
-                            $scope.selectedFunction = filteredSelectedFunction[0];
+                            // update the selected function object because the newly fetched object from the API becomes a different object
+                            var filteredSelectedFunction = $filter('filter')($scope.functions, { name: $scope.selectedFunction.name }, true);
+                            if (filteredSelectedFunction && filteredSelectedFunction.length > 0) {
+                                $scope.selectedFunction = filteredSelectedFunction[0];
+                            } else {
+                                $scope.selectedFunction = undefined;
+                            }
                         } else {
-                            $scope.selectedFunction = undefined;
-                        }
-                    } else {
-                        for (var i = 0; i < $scope.functions.length; i++) {
-                            for (var j = 0; j < response.data.length; j++) {
-                                if ($scope.functions[i].name == response.data[j].name) {
-                                    $scope.functions[i].image = response.data[j].image;
-                                    $scope.functions[i].replicas = response.data[j].replicas;
-                                    $scope.functions[i].invocationCount = response.data[j].invocationCount;
+                            for (var i = 0; i < $scope.functions.length; i++) {
+                                for (var j = 0; j < response.data.length; j++) {
+                                    if ($scope.functions[i].name == response.data[j].name) {
+                                        $scope.functions[i].image = response.data[j].image;
+                                        $scope.functions[i].replicas = response.data[j].replicas;
+                                        $scope.functions[i].invocationCount = response.data[j].invocationCount;
+                                    }
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


gateway/assets/script/bootstrap.js
```diff
@@ -180,29 +180,32 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$interval', '$f
...
-                            // update the selected function object because the newly fetched object from the API becomes a different object
-                            var filteredSelectedFunction = $filter('filter')($scope.functions, {name: $scope.selectedFunction.name}, true);
-                            if (filteredSelectedFunction && filteredSelectedFunction.length > 0) {
-                                $scope.selectedFunction = filteredSelectedFunction[0];
-                            } else {
-                                $scope.selectedFunction = undefined;
-                            }
+            var cl = function (previousItems) {
+                $http.get("../system/functions").then(function (response) {
+                    if (response == null || response.data == null) {
```
kenfdev: I don't think we should change this because this may not catch when the `response` or `response.data` is `undefined`. Also, we should always use triple equals `===` whenever possible (which some existing code doesn't include).

LucasRoesler: I am familiar with the `===` rule in JS, but explicitly chose `==` because it does catch `undefined`.

LucasRoesler:
@kenfdev sorry for the confusion on the review, I should have linked to the logical change directly.  i explicitly split the PR into two commits so that we can separate the logical change from the formatting change

gateway/assets/script/bootstrap.js
kenfdev: Thanks for clearing it out! I wasn't sure how loose it was. 

```
response && response.data
```

The previous statement is probably an anti-pattern in JS if we think strictly, as it catches falsy values (like 0, "").
(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/if...else#Description)

IMHO, I think the previous code was more readable, but am not against this change either.

LucasRoesler: I am ok with rewriting it, but my reason for writing it this way it reduces the nesting / indentation level and keeps the primary logic close to the primary level of the. I generally prefer to check for errors/issues and return early and allow the expected or happy path to be the primary logic of a method

kenfdev: Oh, sorry Lucas, I was missing the whole point. Somehow I only thought the code was changed from no `null` statement to using `null` in the if statement. So I thought the change after that was due to formatting. Sorry about me not understanding the intent.

> I generally prefer to check for errors/issues and return early

This cleared what I thought was strange and I agree about returning early.

alexellis:
Can we do a PR which changes the formatting and nothing else then overlay this PR?

alexellis:
Can we do a PR which changes the formatting and nothing else then overlay this PR?

LucasRoesler:
I already broke it into two commits, if you want I can simply remove the second commit, which is only the formatting changes

alexellis:
Thanks for doing that, but unfortunately it doesn't make it any easier for reviewing on GitHub. Could you do a PR from the first commit?

LucasRoesler:
sure, i figured the two commits could be reviewed individually, but I will roll it back so that only the first is in this PR

gateway/assets/script/bootstrap.js
ivanayov: You can do a check with `(!response && !response.data)` as well.

alexellis: I prefer the original code for this which like Ken says is more readable. Also unless I'm reading this wrong, we reversed the logic. Can you help me understand the change?

```
// do something if there is a value for response.data
if (response && response.data) {
```

Changed to:

```
// do something if and only if either are null?

if (response == null || response.data == null) {
```

LucasRoesler: if there is no response or there is no data, then stop processing the response and return immediately.  Ideally, this would actually throw and error or set an error state in the UI, but that is beyond the scope of this PR

I did this so that

1) the primary intention of the function (update the function data) is the main scope of the function, error and edge cases are checked and dealt with prior to the success case; and
2) it reduces the number of indented blocks and the indentation level as a whole, i find long nested chunks of code hard to read

Issue #470, opened 2018-01-03
Feature Request: CQRS, SAGAs and Event Sourcing as a director function pattern modeled as state machines in yml


dghadi1:
## Expected Behavior
- Model the behavior of [director functions](https://github.com/openfaas/faas/blob/006299c9d848da4071258de9ad1c5d7dae4d37f3/guide/chaining_functions.md) (acting as orchestrators or transactors) as a series of commands, handlers, events and states that can allow for complex enterprise use-cases where handlers can be simple functions combined together by modelling a cohesive business process encompassing multiple steps.

## Context
Orchestrations or transactions when formally modeled gives predictability (through precise design) and maintains behavioral guarantees, when executed in a distributed manner for mission critical use-cases. 
- Retries and predictable recovery from a failed state at each interaction level
- fallbacks for resiliency
- deduping data fetching calls (especially when executing reads in parallel)
- compensating transactions (or rollbacks when mutating) - SAGA
- snapshotting assistance for troubleshooting undesirable behaviors through a distributed event journal
- strict separation of model and code
- strict separation of read/write workloads ([CQRS](http://microservices.io/patterns/data/cqrs.html))
- event driven execution always ([Event Sourcing](http://microservices.io/patterns/data/event-sourcing.html))

- **Orchestrators** are non-side effecting chain of functions (as handlers) and the workflow is modeled as a state-machine using a Command-Handler-Event-State pattern (CQRS). The commands can be any core business event (fetching data) or an agnostic event such as authentication, authorization, introspection, caching, validation etc)
- **Transactors** are side effecting chain of functions (as handlers) and the workflow is modeled as a [SAGA ](http://microservices.io/patterns/data/saga.html) state-machine using a command-Handler-Event-State patter (CQRS). The handlers can either fetch data, or mutate data.
- An **Event Journal** persisting every commands, event, and the state and could be replayed (Event Sourcing pattern)

## Possible Solution
Build a cli command and integrate into the faas-cli by providing a orchestrator-name.yml or a transactor-name.yml and a process to generate the director function image. The orchestrator or transactor can use other function definitions already deployed in faas.

Example model
```
will add a sample model here soon
```
I'll update the issue with more detailed design soon. Any help/direction on how to do this faas natively will be greatly appreciated (using kubernetes).

## Related issues
#91

alexellis:
[...]

PR #997, opened 2018-12-16
Config metrics endpoint

hxalid:
Make metrics endpoint configurable via 'metrics_endpoint'
environment variable. Default is '/metrics'.

Signed-off-by: Khalid Hasanov <xalid.h@gmail.com>

This change is a new feature making it possible to override gateway metrics
endpoint.  

## Description

This change is a new feature making it possible to override gateway metrics
endpoint. By using metrics_endpoint environment variable users can override
gateway metrics endpoints. The default value is '/metrics' which is the same as before.
 
## Motivation and Context

Not all Prometheus setups are configured to scrape '/metrics' endpoint and
therefore, this change would make it easier to have a consistent Prometheus 
config for different systems.

https://github.com/openfaas/faas/issues/998

## How Has This Been Tested?

The change has been used with OpenFaaS Nomad plugin.

The tests were run by overriding 'metrics_endpoint' environment variable and 
making sure Prometheus can scrape the metrics.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/README.md
+++ b/gateway/README.md
@@ -71,6 +71,7 @@ The gateway can be configured through the following environment variables:
 | `faas_nats_port`    | Port for NATS service. Requrired for asynchronous mode |
 | `faas_prometheus_host`         | Host to connect to Prometheus. Default: `"prometheus"` |
 | `faas_promethus_port`         | Port to connect to Prometheus. Default: `9090` |
+| `metrics_endpoint`            | Gateway metrics endpoint scraped by Prometheus. Default: `/metrics` |
 | `direct_functions`            | `true` or `false` -  functions are invoked directly over overlay network by DNS name without passing through the provider |
 | `direct_functions_suffix`     | Provide a DNS suffix for invoking functions directly over overlay network  |
 | `basic_auth`              | Set to `true` or `false` to enable embedded basic auth on the /system and /ui endpoints (recommended) |
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -182,7 +182,7 @@ func main() {
 	}
 
 	metricsHandler := metrics.PrometheusHandler()
-	r.Handle("/metrics", metricsHandler)
+	r.Handle(config.MetricsEndpoint, metricsHandler)
 	r.HandleFunc("/healthz", handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver, nilURLTransformer)).Methods(http.MethodGet)
 
 	r.Handle("/", http.RedirectHandler("/ui/", http.StatusMovedPermanently)).Methods(http.MethodGet)
--- a/gateway/types/readconfig.go
+++ b/gateway/types/readconfig.go
@@ -102,6 +102,12 @@ func (ReadConfig) Read(hasEnv HasEnv) GatewayConfig {
 		cfg.PrometheusHost = prometheusHost
 	}
 
+	metricsEndpoint := hasEnv.Getenv("metrics_endpoint")
+	if len(metricsEndpoint) == 0 {
+		metricsEndpoint = "/metrics"
+	}
+	cfg.MetricsEndpoint = metricsEndpoint
+
 	cfg.DirectFunctions = parseBoolValue(hasEnv.Getenv("direct_functions"))
 	cfg.DirectFunctionsSuffix = hasEnv.Getenv("direct_functions_suffix")
 
@@ -143,6 +149,9 @@ type GatewayConfig struct {
 
 	// Port to connect to Prometheus.
 	PrometheusPort int
+
+	// Metrics endpoint that is going to be scraped by Prometheus
+	MetricsEndpoint string
 
 	// If set to true we will access upstream functions directly rather than through the upstream provider
 	DirectFunctions bool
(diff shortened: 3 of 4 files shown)


alexellis:
Prometheus can be configured which endpoint it should scrape through configuration and through annotations when connected to tools such as Weave Scope. This change would not work for OpenFaaS on Kubernetes since we'd have to change the supporting configuration. OpenFaaS bundles its own Prometheus which is set up specifically to control metrics and scaling in a known way. 

Alex

alexellis:
Derek add label: design/review

alexellis:
Please also see our contribution guide which states you should not work on or submit a PR woehtour an approved issue first.

hxalid:
Sorry not being clear in my previous explanations. This change is needed when there is an existing Prometheus setup that is not configured to scrape the default /metrics endpoint. I appreciate you pointed out links to Weave Scope. However, I'm not sure it would work in Nomad for instance. Also, I'm well aware that Prometheus can be configured to scrape different metrics endpoint, however, it is not convenient when Prometheus needs to be reconfigured for each new system that is being introduced into the existing infrastructure.

alexellis:
In order to use OpenFaaS you will need to configure AlertManager and Prometheus separately. Having this configuration item doesn't make sense for our users at large and carries a cost of maintenance. 

The standard endpoint for Prometheus metrics is `/metrics`. Out of interest what is your system looking for and why?

hxalid:
I totally agree with that. However, from my experience, where there is an existing Prometheus cluster that is configured to use a non-default metrics endpoint (for different reasons) it becomes necessary to override this setting at OpenFaaS.  There could be various reasons not to use a non-default scrape endpoint. For instance, we may have a load balancer in front of an existing Nomad cluster and when we hit the default endpoint it would get load balanced across different Nomad servers. However, it would make more sense if Prometheus scrapes a non-balanced specific endpoint. It is just one reason. If we are talking about a more complex established microservices infrastructure where all microservices by default expose metrics on a non-default endpoint, it would be much easier to integrate OpenFaaS into the existing infrastructure without changing the existing configurations.

I hope it answers to your questions. These all were a reason this PR were created to address the real issue I hit during evaluating OpenFaaS on Nomad.

acornies:
@hxalid Can handle this internal requirement another way? As in, at a prometheus level, use a custom metrics endpoint that you can't get around, then you create a reverse proxy or some other tcp interceptor to forward requests to http(s)://gateway:{port}/metrics in your nomad cluster? This would satisfy both requirements, without optionally changing the signature on the gateway component itself.
I know this is a workaround, but it sounds like workarounds need to be done anyway if strict rules against default metrics endpoints are being enforced in your organization anyway.

hxalid:
@acornies Sure. It is definitely possible to come up with a workaround or build OpenFaaS from source with the required changes. I thought this change would not break anything, as the default value of the endpoint is still /metrics anyway. It should not break anything unless someone explicitly overrides it. Moreover, OpenFaaS already provides over 10 environmental overrides, adding one more override would not break the convention over configuration approach of OpenFaaS. It is something nice to have but is definitely not a blocking thing. I already used a custom built with these changes for my evaluations.

acornies:
I see no functional problems with the change, and if this helps you evaluate OF then that's great! Thanks for opening the PR.

The broader question is support for non-default values of this new option and what implications it has on other projects in the ecosystem. I'm unsure at this point, but I'd lean on @alexellis and other core contributors.

yurilaaziz:
thank you @hxalid, I have the same issue as you. changing /metrics by /health variant (🤣 ). 
even it's bad to not respect standards, I think is better to let users change  the configuration default values. the PR doesn't harm the general behavior. 
LGTM

alexellis:
What is the problem with the health endpoint? This is rarely ever configurable and has knock-on impact.

yurilaaziz:
@alexellis could you explain your point of view of not merging his PR ? 
Even if users raises new issues because they just missed up with the default value of metrics. 
if it needs more refactoring and design we could help. 

Answer to your question about the problem with the health endpoint: there is no problem :p

alexellis:
Hi @hxalid,

Thanks for your detailed blog post.

Did you decide to use OF + Nomad at Jet?

Alex

hxalid:
Hi @alexellis,

Thank you for your feedback on the blog post! 

We decided to wait until the evaluated serverless runtimes provide more IO abstractions and workflow engines. It would probably make sense for us to evaluate serverless runtimes again in near future as serverless technologies evolve quite fast.

It would be great to see a workflow engine and an improved Kafka controller in OpenFaaS. I know you have an open issue (https://github.com/openfaas/faas/issues/470) to address some of these challenges. In terms of Kafka controller, adding batching, retrying, and  potentially being able to use Confluent's Apache Kafka Golang client as an underlying Kafka library for a better performance would be nice improvements.

alexellis:
Great to hear it.

I'll close this PR now, but it would be great if you would like to contribute into the areas you mentioned above.

Thanks to everyone else who commented on this thread, your PRs are more than welcome - please join Slack and we'll help you get started. https://docs.openfaas.com/community. 

Alex

alexellis:
Derek lock

PR #1078, opened 2019-02-04
Remove deploy-stack.armhf.sh script

rdimitrov:
Signed-off-by: Radoslav Dimitrov <rdimitrow@gmail.com>

## Description

Considering that there's now a single `deploy-stack.sh` script, I believe there's no reason to keep a separate one for armhf. 
I understand the reason is to not break any old references, but some time has passed since that change and at the end it might be even rather confusing to keep it.

## Motivation and Context

## Types of changes

## Checklist:

- [ ] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide

### DIFF
--- a/deploy_stack.armhf.sh
+++ b/deploy_stack.armhf.sh
@@ -1,3 +1,0 @@
-#!/bin/sh
-
-sh ./deploy_stack.sh


rgee0:
Not sure about this. The deploy scripts were only consolidated on 25/12 and this will break 2 years worth of content that refers to the armhf version.

alexellis:
I would agree with Richard's point and caution here. 

It will break some blog posts, but probably not _that bad_.. I suspect people will come looking then see this issue or ask and then get pointed at deploy_stack.sh.

For now it seems like a harmless alias. Do we want to agree a self-life for it - so we remove in say 1 or 3 months?

Alex

rdimitrov:
Okay from my side 👍

zeerorg:
Maybe we can collect places where it refers to the `deploy_stack.armhf` version here if we stumble upon it so that the changes are updated

alexellis:
Those places happen to be across dozens of blog posts 😄  If we break it people may just arrive here? I'm not sure I want the pain of dozens of people blaming me for this.

zeerorg:
Ohh, didn't think about the blog posts that are referring to current functionality. We should leave it in that case. we don't want to break personal scripts and blog posts 😛

alexellis:
Let's see what happens. I'll merge it.

Issue #1081, opened 2019-02-07
Proposal: move /metrics to a separate HTTP port


alexellis:
## Task

We should move the /metrics/ path scraped by Prometheus for internal use to a separate HTTP port. This separates concerns.

## Possible Solution

* Figure out the port we want to use (not 8081, since that clashes with the provider)
* Add another HTTP server
* Bind the handler from Prometheus
* Update helm chart, YAML and docker-compose for all platforms.
* Test end to end

viveksyngh:
I would like to work on this one.

alexellis:
You can assign it with Derek

viveksyngh:
Derek assign: me

PR #1085, opened 2019-02-13
Add new HTTPServer to serve /metrics on port 8082

viveksyngh:
This is WIP PR to a new HTTPServer on port 8082 in a goroutine to serve
/metrics endpoint on a different port.

This also update the configurations and compose files.

Fixes: #1081

## Motivation and Context

 #1081

## How Has This Been Tested?

Tested on local swarm cluster:
screen shot 2019-02-15 at 7 37 13 pm

screen shot 2019-02-15 at 7 37 30 pm

## Types of changes

## Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,7 @@ services:
     gateway:
         ports:
             - 8080:8080
+            - 8082:8082
         image: openfaas/gateway:0.11.1
         networks:
             - functions
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,6 @@ services:
     gateway:
         ports:
             - 8080:8080
-            - 8082:8082
         image: openfaas/gateway:0.11.1
         networks:
             - functions
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -207,7 +207,22 @@ func main() {
 	}
 
 	metricsHandler := metrics.PrometheusHandler()
-	r.Handle("/metrics", metricsHandler)
+
+	//Create a new server to serve /metrics endpoint
+	go func() {
+		routes := mux.NewRouter()
+		routes.Handle("/metrics", metricsHandler)
+		port := 8082
+		s := &http.Server{
+			Addr:           fmt.Sprintf(":%d", port),
+			ReadTimeout:    config.ReadTimeout,
+			WriteTimeout:   config.WriteTimeout,
+			MaxHeaderBytes: http.DefaultMaxHeaderBytes,
+			Handler:        routes,
+		}
+		log.Fatal(s.ListenAndServe())
+	}()
+
 	r.HandleFunc("/healthz", handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver, nilURLTransformer)).Methods(http.MethodGet)
 
 	r.Handle("/", http.RedirectHandler("/ui/", http.StatusMovedPermanently)).Methods(http.MethodGet)
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -232,11 +232,11 @@ func runMetricsServer() {
 	metricsHandler := metrics.PrometheusHandler()
 	router := mux.NewRouter()
 	router.Handle("/metrics", metricsHandler)
-	router.HandleFunc("/healthz", healthzHandler)
+	router.HandleFunc("/healthz", handlers.HealthzHandler)
 
 	port := 8082
-	readTimeout := time.Duration(5) * time.Second
-	writeTimeout := time.Duration(5) * time.Second
+	readTimeout := 5 * time.Second
+	writeTimeout := 5 * time.Second
 
 	s := &http.Server{
 		Addr:           fmt.Sprintf(":%d", port),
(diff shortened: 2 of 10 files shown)


gateway/server.go
```diff
@@ -207,7 +207,22 @@ func main() {
...
+		routes := mux.NewRouter()
+		routes.Handle("/metrics", metricsHandler)
+		port := 8082
+		s := &http.Server{
+			Addr:           fmt.Sprintf(":%d", port),
+			ReadTimeout:    config.ReadTimeout,
+			WriteTimeout:   config.WriteTimeout,
+			MaxHeaderBytes: http.DefaultMaxHeaderBytes,
+			Handler:        routes,
+		}
```
LucasRoesler: How do you plan on stopping this goroutine when the main thread stops? The server should be defined outside of the goroutine, there should be some kind of signal catch, and then sending `s.Shutdown()` to stop this gracefully.

viveksyngh: I think when main thread stops it will also stop the go routines launched from main thread, it may not do the cleanup on the resources used.  I will implement the server to shutdown gracefully.

alexellis:
Graceful shutdown should first be implemented separately for the main server - raise an issue if you want? I'm OK for this metrics endpoint to be shutdown from SIGTERM in the initial version.

alexellis:
Do we need a /healthz on the new port, too?

gateway/server.go
alexellis: I think we could set a shorter, static timeout value for the HTTP server given that this is not serving function requests.

gateway/server.go
```diff
@@ -207,7 +207,22 @@ func main() {
 	}
 
 	metricsHandler := metrics.PrometheusHandler()
-	r.Handle("/metrics", metricsHandler)
+
+	//Create a new server to serve /metrics endpoint
```
alexellis:
```
	// Listen on a separate HTTP port for Prometheus metrics to keep this accessible from 
        // the internal network only. 
```

gateway/server.go
```diff
@@ -207,7 +207,22 @@ func main() {
...
+	go func() {
```
alexellis: Let's do this in a separate method at least?

docker-compose.yml
```diff
@@ -3,6 +3,7 @@ services:
     gateway:
         ports:
             - 8080:8080
+            - 8082:8082
```
alexellis: Does this work if you remove the port? It should still be accessible from within the cluster. We don't need to expose it generally.

prometheus/prometheus.yml
```diff
@@ -32,7 +32,7 @@ scrape_configs:
     scrape_interval: 5s
     dns_sd_configs:
       - names: ['tasks.gateway']
-        port: 8080
+        port: 8082
```
alexellis: From an operational standpoint I can't merge this PR until we do a new release of the gateway, this needs to be two separate PRs.

viveksyngh: Okay .. so these prometheus configs should be put in a different PR and one PR to add the gateway changes ?

docker-compose.yml
viveksyngh: Yes, It does and I have removed `8082` from exposed port.

gateway/server.go
viveksyngh: I have put this in a separate method and added `/healthz` endpoint.

gateway/server.go
viveksyngh: Done.

gateway/server.go
viveksyngh: I have update the server to use `30s` for `readTimeout` and `writeTimeout`

alexellis:
Do we need 30s for the metrics endpoint? It seems excessive

viveksyngh:
how about `5s` ?

alexellis:
That sounds more like what I had in mind

gateway/server.go
alexellis: `30s` seems too high for this purpose, let's reduce it to `5s`?

prometheus/prometheus.yml
alexellis: Yes

gateway/server.go
```diff
@@ -224,3 +225,39 @@ func main() {
...
+		ReadTimeout:    readTimeout,
+		WriteTimeout:   writeTimeout,
+		MaxHeaderBytes: http.DefaultMaxHeaderBytes,
+		Handler:        router,
+	}
+
+	log.Fatal(s.ListenAndServe())
+}
+
+func healthzHandler(w http.ResponseWriter, r *http.Request) {
```
alexellis: Do we already have a healthz handler for the gateway itself? Perhaps this handler should be moved to the `handlers` package or shared?

gateway/server.go
```diff
@@ -224,3 +225,39 @@ func main() {
...
+//runMetricsServer Listen on a separate HTTP port for Prometheus metrics to keep this accessible from
+// the internal network only.
+func runMetricsServer() {
+	metricsHandler := metrics.PrometheusHandler()
+	router := mux.NewRouter()
+	router.Handle("/metrics", metricsHandler)
+	router.HandleFunc("/healthz", healthzHandler)
+
+	port := 8082
+	readTimeout := time.Duration(5) * time.Second
```
alexellis: Does `5 * time.Second` not work?

gateway/server.go
viveksyngh: gateway using reverse proxy handler for `healthzHandler` as well. Shouldn't this have different healthz handler?

gateway/server.go
viveksyngh: That will also work.

burtonr:
@viveksyngh This PR looks to be very nearly complete. The only thing preventing it from merging (from what I can tell) is removing the change to the `prometheus/*.yml` files.

viveksyngh:
@burtonr I have removed prometheus changes. Will raise a different PR for that.

PR #1165, opened 2019-04-14
Bundle watchdog binaries into an image

### LINKED ISSUE #969: Proposal: Publish Docker image(s) with watchdog binary
## Description

We currently need to install curl / wget to fetch the watchdog / of-watchdog binary for templates. What if we published a set of docker images with each watchdog binary per platform - or one larger docker image with all inside and then used that to speed up the build?

```
FROM openfaas/watchdog:0.9.11 as watchdog
FROM go:1.1.1
COPY --from watchdog /fwatchdog /usr/bin/watchdog
```

Our images would be easier to maintain as per above vs:

```
FROM golang:1.10.4-alpine3.8 as builder

RUN apk --no-cache add curl \
    && echo "Pulling watchdog binary from Github." \
    && curl -sSL https://github.com/openfaas/faas/releases/download/0.9.6/fwatchdog > /usr/bin/fwatchdog \
    && chmod +x /usr/bin/fwatchdog \
    && apk del curl --no-cache
```

What do we think? Perhaps it might help with #957 via @rdimitrov

rgee0:
Signed-off-by: Richard Gee <richard@technologee.co.uk>

## Description
Adds a packager Dockerfile to create an image that contains the fwatchdog binaries.

## Motivation and Context
Fixes #969 

## How Has This Been Tested?
Successfully ran locally.
CI passed

## Types of changes

## Checklist:

### DIFF
--- a/watchdog/Dockerfile.packager
+++ b/watchdog/Dockerfile.packager
@@ -1,7 +1,6 @@
 FROM openfaas/watchdog:build as build
 FROM scratch
 
-COPY --from=build /go/src/github.com/openfaas/faas/watchdog/watchdog ./fwatchdog
-COPY --from=build /go/src/github.com/openfaas/faas/watchdog/watchdog-armhf ./fwatchdog-armhf
-COPY --from=build /go/src/github.com/openfaas/faas/watchdog/watchdog-arm64 ./fwatchdog-arm64
-COPY --from=build /go/src/github.com/openfaas/faas/watchdog/watchdog.exe ./fwatchdog.exe
+ARG PLATFORM
+
+COPY --from=build /go/src/github.com/openfaas/faas/watchdog/watchdog$PLATFORM ./fwatchdog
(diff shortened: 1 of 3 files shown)


alexellis:
This is definitely useful. What I'm trying to get to is where we have a different tagged image for each architecture/binary. It would mean that rather than using curl we could use multistage builds. 

If all of the binaries are in a single image then that could be quite a big download. I think this can be done on a cloud VM even for arm. Maybe by using scratch?

rgee0:
Currently we cross compile for all platforms in CI - I'm not sure what the reference to a cloud VM pertains to but it feels like you're suggesting we take the image creation out of CI? 

The image is 25MB

watchdog/Dockerfile.packager
```diff
@@ -0,0 +1,6 @@
+FROM openfaas/watchdog:build as build
+FROM scratch
+
+ARG HARDWARE
```
alexellis: Minor nit - since Windows isn't "HARDWARE" - platform may work better. This doesn't prevent a merge though.

alexellis:
This is what I had in mind :+1:

alexellis:
Thank you @rgee0

PR #1210, opened 2019-06-05
Add basic auth plugin

alexellis:
Signed-off-by: Alex Ellis <alexellis2@gmail.com>

## Description

This plugin can be tested as a stand-alone Golang app, or
deployed as a Docker image with the following image:
openfaas/basic-auth-plugin:0.1.0

## Motivation and Context

- [x] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

#1209 

## How Has This Been Tested?

```
echo -n alex > basic-auth-user
echo -n password > basic-auth-password

go build && secret_mount_path="`pwd`" port=8088 ./basic-auth 
2019/06/05 12:06:12 Listening on: 8088

curl http://alex:admin@localhost:8088/validate -i
HTTP/1.1 200 OK
Date: Wed, 05 Jun 2019 11:06:51 GMT
Content-Length: 0

curl http://not-alex:admin@localhost:8088/validate -i
HTTP/1.1 401 Unauthorized
Date: Wed, 05 Jun 2019 11:07:03 GMT
Content-Length: 19
Content-Type: text/plain; charset=utf-8

2019/06/05 12:06:51 Validated request 200.
2019/06/05 12:07:03 Validated request 401.

```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] I have signed-off my commits with `git commit -s`
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #1209, opened 2019-06-05
Feature: Enable nginx-style pluggable authentication


alexellis:
Feature: Enable nginx-style pluggable authentication

## Expected Behaviour

A user should be able to specify either a function endpoint or a micro-service to authenticate requests through the API Gateway.

This would extend beyond the current basic-auth validation.

This would be similar to the way Nginx implements external auth:

> NGINX and NGINX Plus can authenticate each request to your website with an external server or service. To perform authentication, NGINX makes an HTTP subrequest to an external server where the subrequest is verified. If the subrequest returns a 2xx response code, the access is allowed, if it returns 401 or 403, the access is denied. Such type of authentication allows implementing various authentication schemes, such as multifactor authentication, or allows implementing LDAP or OAuth authentication.

https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-subrequest-authentication/

## Current Behaviour

Basic auth is the only option, or fronting the gateway with another proxy / gateway such as Nginx, Ambassador, or Kong.

## Possible Solution

* Add configuration options to the gateway
* Update handlers that add in-process basic auth to add a "external delegating HTTP handler"
* Update the default configuration
* Create an example basic-auth validation service

We need to consider the scope of the validation. Is this just for communication to the gateway, or for all internal communication too?

## Scope

Internal communication between system-components would continue to use the system account (basic-auth), but external user auth can now be controlled by "user-auth".

"user-auth" for the first auth plugin happens to be the same as basic-auth, it could use a separate secret.

![pluggable-auth](https://user-images.githubusercontent.com/6358735/58979842-dd71ae00-87c6-11e9-9f55-2fdcf1560b2b.png)

The initial version will only cover system APIs between the user and the gateway, but going forward we do also have the scope to apply user-auth to public functions.

Phase 1: /system /ui are covered by the “user auth” instead of built-in auth
Phase 2: optionally extends user-auth to /function/

alexellis:
Derek close: released and available from 0.13.9 and onwards

PR #1211, opened 2019-06-05
Add external auth handler and gateway config options

alexellis:

## Description

This commit adds an external auth handler which can be used to
wrap existing handlers, so that they delegate their requests
to an upstream URL before allowing a request to pass through
to an upstream API.

Gateway config also gains two new options:

* auth_proxy_url
* auth_proxy_pass_body

## Motivation and Context

#1209 #1210 

## How Has This Been Tested?

The new handler has been tested with unit tests for positive / negative cases

## Types of changes

## Checklist:

### DIFF
--- a/gateway/handlers/external_auth.go
+++ b/gateway/handlers/external_auth.go
@@ -1,0 +1,1 @@
+package handlers
--- a/gateway/handlers/external_auth.go
+++ b/gateway/handlers/external_auth.go
@@ -11,9 +11,9 @@ func MakeExternalAuthHandler(next http.HandlerFunc, upstreamTimeout time.Duratio
 	return func(w http.ResponseWriter, r *http.Request) {
 		req, _ := http.NewRequest(http.MethodGet, upstreamURL, nil)
 
-		deadlineContext, cancel := context.WithDeadline(
+		deadlineContext, cancel := context.WithTimeout(
 			context.Background(),
-			time.Now().Add(upstreamTimeout))
+			upstreamTimeout)
 
 		defer cancel()
 
--- a/gateway/handlers/external_auth.go
+++ b/gateway/handlers/external_auth.go
@@ -1,1 +1,37 @@
 package handlers
+
+import (
+	"context"
+	"net/http"
+	"time"
+)
+
+// MakeExternalAuthHandler make an authentication proxy handler
+func MakeExternalAuthHandler(next http.HandlerFunc, upstreamTimeout time.Duration, upstreamURL string, passBody bool) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		req, _ := http.NewRequest(http.MethodGet, upstreamURL, nil)
+
+		deadlineContext, cancel := context.WithDeadline(
+			context.Background(),
+			time.Now().Add(upstreamTimeout))
+
+		defer cancel()
+
+		res, err := http.DefaultClient.Do(req.WithContext(deadlineContext))
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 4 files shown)


gateway/handlers/external_auth_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+		res, err := http.DefaultClient.Do(req)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+		if res.Body != nil {
+			defer res.Body.Close()
+		}
+
+		if res.StatusCode == http.StatusOK {
```
LucasRoesler: It can also be good to allow non-200 response from the auth request to be returned directly to the user.  There can be headers or content in the body that enable or are required for the auth flow. 

nginx does this only for the 401 adn 403 responses http://nginx.org/en/docs/http/ngx_http_auth_request_module.html

Traefik returns all non-200 responses from the auth request https://docs.traefik.io/configuration/entrypoints/#forward-authentication

gateway/handlers/external_auth_test.go
```diff
@@ -0,0 +1,72 @@
...
+	if rr.Code != want {
+		t.Errorf("Status incorrect, want: %d, but got %d", want, rr.Code)
+	}
+}
+
+func MakeExternalAuthHandler(next http.HandlerFunc, upstreamURL string, passBody bool) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		req, _ := http.NewRequest(http.MethodGet, upstreamURL, nil)
+
+		res, err := http.DefaultClient.Do(req)
```
stefanprodan: Maybe use a timeout context of 5s (or some a lower value)

gateway/handlers/external_auth_test.go
```diff
@@ -0,0 +1,72 @@
...
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
```
stefanprodan: Shouldn't we break execution here?

gateway/handlers/external_auth_test.go
alexellis: I agree, I think it would be OK to add it after the first batch of PRs with an additional set of tests for that. We have some convenient 2xx checking-code somewhere we can re-use.

gateway/handlers/external_auth_test.go
alexellis: Yes

gateway/handlers/external_auth_test.go
alexellis: Added, please review.

gateway/handlers/external_auth.go
```diff
@@ -0,0 +1,37 @@
...
+	"net/http"
+	"time"
+)
+
+// MakeExternalAuthHandler make an authentication proxy handler
+func MakeExternalAuthHandler(next http.HandlerFunc, upstreamTimeout time.Duration, upstreamURL string, passBody bool) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		req, _ := http.NewRequest(http.MethodGet, upstreamURL, nil)
+
+		deadlineContext, cancel := context.WithDeadline(
```
stefanprodan: You could use `context.WithTimeout(context.Background(), upstreamTimeout)`

gateway/handlers/external_auth_test.go
LucasRoesler: sure

gateway/handlers/external_auth.go
alexellis: That's a good point.

PR #1213, opened 2019-06-07
Switch gateway to external auth

alexellis:
Switch gateway to external auth

## Description

This commit moves the OpenFaaS gateway from using in-process
basic-auth for everything to use an external auth URL instead.

When auth is not enable, this functionality is not added to the
handlers and behaves as before. When enabled, the configured
plugin with authenticate requests.

## Motivation and Context

#1209 

## How Has This Been Tested?

Tested e2e on Docker Swarm with positive and negative tests.

## Types of changes

## Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ services:
     gateway:
         ports:
             - 8080:8080
-        image: openfaas/gateway:0.13.6
+        image: openfaas/gateway:0.13.7-rc2
         networks:
             - functions
         environment:
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -83,6 +83,8 @@ func main() {
 		functionURLTransformer = nilURLTransformer
 	}
 
+	decorateExternalAuth := handlers.MakeExternalAuthHandler
+
 	faasHandlers.Proxy = handlers.MakeForwardingProxyHandler(reverseProxy, functionNotifiers, functionURLResolver, functionURLTransformer)
 
 	faasHandlers.RoutelessProxy = handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver, nilURLTransformer)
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -3,7 +3,7 @@ services:
     gateway:
         ports:
             - 8080:8080
-        image: openfaas/gateway:0.13.6
+        image: openfaas/gateway:0.13.7-rc2
```
LucasRoesler: You should be able to remove the `basic_auth: "${BASIC_AUTH:-true}"` env variable now, right?

alexellis:
I think that we need a flag to turn on/off in YAML but we could change the name.

gateway/server.go
```diff
@@ -131,25 +133,25 @@ func main() {
 
 	if credentials != nil {
```
LucasRoesler: It seems odd for this to be based on the credentials, the credentials shouldn't be used in the gateway anymore and are specific to only one type of auth.  It seems like the check for if there should be auth should be handled by the value for `config.AuthProxyURL` and the parsing of the credentials should be removed

alexellis: Please see the diagram in the attached issue.

The secret is used for core-service commutation. The plugin is used for user-auth to the gateway.

LucasRoesler: There seems to only be one auth configuration, via the `MakeExternalAuthHandler`. How can the internal service communication use basic-auth if the auth proxy points at an external authentication that is not basic auth?  This works right now, because the only external proxy is the basic auth verification but it seems like it would break as soon as you tried to validate with any other external auth proxy.

alexellis: You are correct. I think the upstreamProxy needs to have the credentials for the upstream calls to authenticated endpoints.

docker-compose.yml
alexellis: I will check

gateway/server.go
alexellis: Function proxy would remain as it is:

```
	faasHandlers.Proxy = handlers.MakeForwardingProxyHandler(reverseProxy, functionNotifiers, functionURLResolver, functionURLTransformer)
```

List for instance, would take a new function override so that it could provide relevant headers for calls.. 

```
func applyServiceAuth(r *http.Request) {
 r.Header.SetBasicAuth(user, pass)
}

	faasHandlers.ListFunctions = handlers.MakeForwardingProxyHandler(reverseProxy, forwardingNotifiers, urlResolver, nilURLTransformer, applyServiceAuth)
```

alexellis: Please see: https://github.com/openfaas/faas/pull/1214

alexellis:
Derek close: in favour of #1214

PR #1325, opened 2019-09-30
Fix event order

jmickey:
Signed-off-by: Josh Michielsen <github@mickey.dev>


alexellis:
I thought the event by Keiran / Affix / CLI was out of sync too and the one next to it

jmickey:
@alexellis should be fixed now.

PR #1376, opened 2019-11-11
Add vCenter Event Broker Appliance to projects

embano1:
The vCenter Event Broker Appliance (VMware Fling) uses OpenFaaS to
trigger functions based on vSphere events.

Signed-off-by: Michael Gasch <mgasch@vmware.com>


alexellis:
Hi Michael, thanks for this PR (which is fine), but I was hoping to get the event where you spoke at VMworld listed in the "events 2019" section?

embano1:
Sure, I just don't see it listed in the replays. If EU was not recorded, we'll use the US version (which also discusses OpenFaaS but is before the actual release of the appliance).

embano1:
Updated with the recorded session video @alexellis

PR #1382, opened 2019-11-21
Update ADOPTERS.md

goncalo-oliveira:
Include Fonix Telematics in the list of OpenFaaS adopters.

Signed-off-by: Goncalo Oliveira <goncalo.oliveira@fonixtelematics.com>


alexellis:
Merging, thank you for getting in touch.

Please think about whether you can write a little more about the application/use-case?

PR #1403, opened 2019-12-14
WIP: Update faas-swarm version in docker-compose files

Waterdrips:
Signed-off-by: Alistair Hey <alistair@heyal.co.uk>

## Description

Assuming that the PR here : https://github.com/openfaas/faas-swarm/pull/63 goes in, and the next faas-swarm container version is 0.8.3, this pr will update the swam deployment to use this new version

This is WIP on waiting for the new container versions


alexellis:
Needs to be `0.8.2`, but the arm builds are manual, so we'll update those later.

PR #1409, opened 2019-12-17
Change how and when we fetch and parse namespace info

### LINKED ISSUE #1408: Namespace selector does not update the function list
## My actions before raising this issue
- [ ] Followed the [troubleshooting guide](https://docs.openfaas.com/deployment/troubleshooting/)
- [ ] Read/searched [the docs](https://docs.openfaas.com/)
- [x] Searched [past issues](/issues)

If I  deploy functions to the default namespace and to an additional namespace, the namespace selector in the gateway UI does not appear to work.  The API returns the expected result with a new function list, but he UI does not update to reflect this

In this screenshot you can see the latest API response but with the wrong function list rendered
image

## Expected Behaviour

The function list should change to show the functions from the selected namespace

## Current Behaviour

Found while walking through a demo of multinamespace support

1. Deploy on kubernetes with the latest helm chart `openfaas-5.3.8`, make sure to set `clusterRole` to true
2. Create an annotate an extra namespace
    ```sh
    kubectl create ns dev-fn
    kubectl annotate ns dev-fn openfaas="true"
   ```
2. use the cli to deploy left pad to an extra namespace, `faas-cli store deploy Left-Pad -n dev-fn`
3. also deploy a function to the default namespace `faas-cli store deploy NodeInfo`
3. Open the gateway ui
5. use the namespace selector to change to `dev-fn`
6. observer that the function list does not change, it should still just show `NodeInfo`

## Context

Relates to #1379

## Your Environment

faas-netes

* Operating System and version (e.g. Linux, Windows, MacOS):
osx

Waterdrips:
## Description

The initial namespace select implementation had some challenges, especially around non-default installs (no openfaas-fn namespace)
This has removed lots of those issues and added some more user friendly bits too (defaulting to first returned namespace from backend, passing that namespace to the function-create as default)

Signed-off-by: Alistair Hey <alistair@heyal.co.uk>

## Motivation and Context

closes #1408 
closes #1408 

## How Has This Been Tested?

kubernetes:
Install with 4 active namespaces, switching works, adding a fn works (for each namespace)
Can add the same fn to each namespace
Can call each fn, see it calls correct one in logs
Can delete functions

Swarm:
No NS selector, can create, call and delete fns.

## Types of changes

## Checklist:

### DIFF
--- a/gateway/assets/script/bootstrap.js
+++ b/gateway/assets/script/bootstrap.js
@@ -17,13 +17,14 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$interval', '$f
         $scope.invocationStatus = "";
         $scope.invocationStart = new Date().getTime();
         $scope.roundTripDuration = "";
-        $scope.selectedNamespace = "openfaas-fn";
-        $scope.allNamespaces = ["openfaas-fn"];
+        $scope.selectedNamespace = "";
+        $scope.allNamespaces = [""];
         $scope.invocation = {
             contentType: "text"
         };
 
         $scope.baseUrl = $location.absUrl().replace(/\ui\/$/, '');
+
         try {
             $scope.canCopyToClipboard = document.queryCommandSupported('copy');
         } catch (err) {
@@ -65,13 +66,17 @@ app.controller("home", ['$scope', '$log', '$http', '$location', '$interval', '$f
         var fetchFunctionsDelay = 3500;
         var queryFunctionDelay = 2500;
 
-        var fetchNamespacesInterval = function() {
-            $http.get("../system/namespaces")
-                .then(function(response) {
-                    $scope.allNamespaces = response.data;
-                })
-        };
 
+
+
+        $http.get("../system/namespaces")
+            .then(function(response) {
+                $scope.allNamespaces = response.data;
+                if ($scope.selectedNamespace === "" && response.data && response.data[0] && response.data[0] !== "") {
+                    $scope.selectedNamespace = response.data[0]
+                    refreshData($scope.selectedNamespace)
+                }
+            })
 
 
         var fetchFunctionsInterval = $interval(function() {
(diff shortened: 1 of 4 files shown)


gateway/assets/script/bootstrap.js
```diff
@@ -603,4 +600,8 @@ function uuidv4() {
     return ([1e7]+-1e3+-4e3+-8e3+-1e11).replace(/[018]/g, function(c) {
         return (c ^ cryptoInstance.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16)
     })
-}  
\ No newline at end of file
+}
+
+function getNamespaceSuffix(namespace) {
```
LucasRoesler: this is a small complaint, but it looks like this returns a query parameter, not a suffix.  It also requires this method to be called first because it returns the `?`, this will make it a little fragile is someone doesn't recognize that the order is important. Javascript has a URL object https://developer.mozilla.org/en-US/docs/Web/API/URL that might be easier to use and more robust to work with, since the search params object has a set method https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/set

Waterdrips: @LucasRoesler 
The URL object doesn't support relative paths, and the CORS blocking on localhost means that using `new URL(document.lcoation.domain)` to generate a url `localhost:8080` and then using the url object to do this stuff doesnt work because (at least) chrome block the requests (when i tried anyway) a rename and re-structure of this is incoming. Hopefully its better.

LucasRoesler: wow that is an incredibly disappointing API in JS then :(

Waterdrips: I mean it worked, but because mostly people will be using localhost CORS gets in the way and makes things hard

alexellis:
Merging based upon review from @LucasRoesler

Issue #776, opened 2018-07-18
End-user community / ADOPTERS.md


alexellis:
If your company has adopted OpenFaaS then you can add your use-case to the ADOPTERS.md file and/or request your logo to be listed. All logos listed must have an entry in the ADOPTERS.md file.

You will need to request authorisation from your manager and employer.

## Process

1) Send a PR to https://github.com/openfaas/faas/blob/master/ADOPTERS.md with a brief description of how you're using OpenFaaS, this should include your use-case
2) Individual sponsorships can be arranged via GitHub Sponsors - if you're implementing or managing OpenFaaS, then you should subscribe to get the newsletters on updates, security patches and features - https://insiders.openfaas.io/
3) Corporations that use OpenFaaS in production should contribute towards its upkeep, whilst the software may be used under the terms of the MIT License, it requires full-time engineering. Pick a tier - bronze/gold/platinum - https://github.com/openfaas/faas/blob/master/BACKERS.md
4) By sponsoring your logo is associated with one of the fastest growing Open Source developer communities, you also reduce the risk of the project going away due to lack of funds
5) Email alex@openfaas.com with written permission for the logo to be listed and comment on this thread if you like.

Asset for logo: we will need a URL to a svg or transparent png logo, a text title and a company URL.

Logos are shown at https://www.openfaas.com/

alexellis:
Other inspiration from the CNCF End User Community - https://www.cncf.io/people/end-user-community/

LucasRoesler:
Contiamo is using OpenFaaS and would like to be included on the page.  What format, shape, and size should the logo be?

alexellis:
It would be ideal if it was a similar format / shape to the ones on the CNCF page linked above.

rimiti:
Hello @alexellis !

Dim Solution is using OpenFaas (soon in production), you can use our logo: 
- [Website](https://dimsolution.com/)
- [Black logo (prefered)](https://dimsolution.com/wp-content/uploads/2018/03/logo-black-1.png)
- [White logo](https://dimsolution.com/wp-content/uploads/2018/03/logo-white.png)

Regards,

ewilde:
@alexellis 

Form3

* [website link: https://form3.tech](https://form3.tech)
* [logo designed for white backgrounds](https://form3.tech/images/press-media/Form3FinancialCloud_Logo.svg)
* [logo designed for dark backgrounds](https://form3.tech/images/press-media/Form3FinancialCloud_Logo_white.svg)

[...]

PR #1411, opened 2019-12-19
Record metrics for invocations when they start

alexellis:

## Description

Record metrics for invocations when they start

## Motivation and Context

* This experimental patch records metrics as invocations start so that the metrics can be used to make better scale to zero decisions in faas-idler.

## How Has This Been Tested?

Tested with Kubernetes on a single-node cluster, metrics
reported as expected. Existing metrics still report.

## Types of changes


gateway/metrics/metrics.go
```diff
@@ -81,6 +82,15 @@ func BuildMetricsOptions() MetricOptions {
 		[]string{"method", "path", "status"},
 	)
 
+	startedCounter := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: "function",
```
LucasRoesler: I think the opts are more correct if we use 

```go
prometheus.CounterOpts{
    Namespace: "function",
    Subsystem: "requests",
    Name: "started_total",
    Help: "The total number of function HTTP requests started.",
}
```

It should result in the same metric: `function_requests_started_total` per https://godoc.org/github.com/prometheus/client_golang/prometheus#Opts

gateway/handlers/notifiers.go
```diff
@@ -46,19 +46,26 @@ type PrometheusFunctionNotifier struct {
...
+			Observe(seconds)
+
+		code := strconv.Itoa(statusCode)
+
+		p.Metrics.GatewayFunctionInvocation.
+			With(prometheus.Labels{"function_name": serviceName, "code": code}).
+			Inc()
+	} else if event == "started" {
+		serviceName := getServiceName(originalURL)
+		p.Metrics.StartedCounter.WithLabelValues(serviceName).Inc()
```
LucasRoesler: I think it would be nice if the metric name was symmetric with the metric that capture the end of the request `GatewayFunctionInvocation` and  `GatewayFunctionInvocationStarted`  or using `FunctionInvocationCompleted` and `FunctionInvocationStarted` to be even more precise

gateway/handlers/notifiers.go
```diff
@@ -46,19 +46,26 @@ type PrometheusFunctionNotifier struct {
 }
 
 // Notify records metrics in Prometheus
-func (p PrometheusFunctionNotifier) Notify(method string, URL string, originalURL string, statusCode int, duration time.Duration) {
-	seconds := duration.Seconds()
-	serviceName := getServiceName(originalURL)
+func (p PrometheusFunctionNotifier) Notify(method string, URL string, originalURL string, statusCode int, event string, duration time.Duration) {
```
LucasRoesler: Can we use an enum for the `event` instead of an arbitrary string? it will be more type safe, easier to read in the docs and IDEs, and we can avoid bugs to do typos.  Especially if we pair it with switch below on the enum constants.

```go
type FunctionEvent int
const (
   FunctionStarted = iota
   FunctionEnded
)
func (d FunctionEvent) String() string {
    return [...]string{"started", "ended"}[d]
}
```

gateway/metrics/metrics.go
```diff
@@ -81,6 +82,15 @@ func BuildMetricsOptions() MetricOptions {
 		[]string{"method", "path", "status"},
 	)
 
+	startedCounter := prometheus.NewCounterVec(
```
LucasRoesler: One more comment, using a standard prometheus metric will still result in lag on this metric and a possible race with the idler between scrapes

alexellis:
Thanks for looking at the metrics. Happy to do polishing after we get to something that works.

gateway/handlers/notifiers.go
alexellis: I don't disagree. This is an experimental PR to see if we can solve at least one of the issues for the user who is in production experiencing this problem today

gateway/handlers/notifiers.go
alexellis: We already capture the end of the request using the existing metric.

What do you suggest? Renaming would be a breaking change, recording double the information would be hideous.

gateway/metrics/metrics.go
alexellis: Agreed, but it will be better than our current worst case.

gateway/metrics/metrics.go
alexellis: @stefanprodan what would you say? I'm happy to take a recommendation as long as its consistent with what we have. Does namespace/subsystem/name have an effect on how metrics are stored? What did we do with the previous metrics?

alexellis:
@stefanprodan @LucasRoesler I've added comments for this, also there was interest from a large enterprise user in this feature to know when long-running requests are started.

gateway/handlers/notifiers.go
LucasRoesler: I am specifically talking about the variable name, `Metrics.StartedCounter` isn't very descriptive.  It is entirely possible we will want other started counters later

gateway/metrics/metrics.go
LucasRoesler: it only impacts how the function is named, it basically treats namespace, subsystem, and name as parts that are joined together.  This just uses the options more semantically and improves the long term readability, imo.  For example, if we wanted to add a counter for sync requests it would then look like 

```go
prometheus.CounterOpts{
    Namespace: "function",
    Subsystem: "async_requests",
    Name: "started_total",
    Help: "The total number of async function HTTP requests started.",
}
```
it should be easier to verify things like spelling mistakes etc because each individual string is shorter

gateway/handlers/notifiers.go
LucasRoesler: ¯\_(ツ)_/¯ if you are happy with it, then i guess we can leave it

StorkZ:
@alexellis @stefanprodan @LucasRoesler 
Hi guys! This fix would help me a lot. I can see that the progress has stalled for 2 weeks now. 
Let me know if there is something I can do to help.

alexellis:
Hi @StorkZ,

Can you introduce who you are, what company you would be using this for and why you feel it's important?

Why do you think that "progress has stalled"? Perhaps this isn't a priority for our roadmap?

Alex

StorkZ:
I work at a saas startup called Kadromierz. We use this project to run exports for clients, that often are required only once a month. We use faas-idler to salce them down to 0 instances. The only problem is that there is a chance, that instance will be scaled to 0 before the function even ran. Currently we handle this by retrying given function call, but it is not a final solution. 
  
I didn't mean it to sound like I'm hastening you. I'm very grateful for this project and if there is something I can do to push this forward, then just let me know.

alexellis:
I'm glad that OpenFaaS is useful to your business, thank you for giving some context. Perhaps you could start off with sending your use-case to the ADOPTERS file? https://github.com/openfaas/faas/issues/776

OpenFaaS is provided for free, features can be prioritized for clients who have a support relationship, otherwise it depends on the roadmap and what little time I can donate to the project.

I'd suggest reading up on how the project works here: [Setting expectations, support and SLAs](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#setting-expectations-support-and-slas)

In the meantime, I'll note that you are interested in this feature. Would you be willing to help with testing?

StorkZ:
Thanks for being patient with me. 
If there is any testing needed just let me know.

gateway/metrics/metrics.go
alexellis: What if we had one counter and then added a label like async: true/false or 1/0?

gateway/handlers/notifiers.go
alexellis: Give me a suggestion

alexellis:
Can you outline a scenario that you are running / running into that we can use to reproduce the behaviour and then test a fix, if we proceed with this patch?

alexellis:
@StorkZ

gateway/metrics/metrics.go
alexellis: Does the suggestion give a consistent name? Wouldn't we want to match with this?

```
	gatewayFunctionInvocation := prometheus.NewCounterVec(
		prometheus.CounterOpts{
			Name: "gateway_function_invocation_total",
			Help: "Individual function metrics",
		},
		[]string{"function_name", "code"},
	)
```

i.e. 

```
prometheus.CounterOpts{
    Namespace: "gateway",
    Subsystem: "function",
    Name: "started_total",
...
```

?

I might be missing something.

gateway/handlers/notifiers.go
alexellis: `gateway_function_invocation_started`?

gateway/handlers/notifiers.go
alexellis: Seems like a reasonable enhancement

gateway/metrics/metrics.go
LucasRoesler: adding another label makes sense to me, especially since it only has two values, it is equivalent to creating two metrics anyway.

LucasRoesler: To your point about 

```go
	gatewayFunctionInvocation := prometheus.NewCounterVec(
		prometheus.CounterOpts{
			Name: "gateway_function_invocation_total",
			Help: "Individual function metrics",
		},
		[]string{"function_name", "code"},
	)
```

I don't think the current metric in this diff has a gateway prefix does it?  I expect code in the current diff to produce `function_requests_started_total`

gateway/handlers/notifiers.go
LucasRoesler: I mean the name of the Go variable and my suggestion was/is `GatewayFunctionInvocationStarted` or `FunctionInvocationStarted`,  so that this code would look like 

```go
p.Metrics.FunctionInvocationStarted.WithLabelValues(serviceName).Inc()
```

gateway/metrics/metrics.go
alexellis: It does have the namespace yes, maybe it changed since you looked at the diff last?

Screenshot 2020-03-12 at 10 45 38

gateway/handlers/notifiers.go
alexellis: I've updated as per that suggestion.

alexellis:
Squashing changes from review into a single commit.

gateway/metrics/metrics.go
LucasRoesler: yeah, looks good. The github UI seems to be wonky for me today.  It seems to be stateful with commit selections sometimes

gateway/handlers/notifiers.go
LucasRoesler: The diff still looks like this

image

alexellis: Ok I see what you mean now. I haven't changed that internal variable, I'll do it next.

alexellis: Updated and retrofitted into the other counter

alexellis:
@StorkZ there are testing instructions and images for this and faasIdler shared in #contributors on Slack.

Issue #1422, opened 2020-01-05
Http_Content_Length can't be negative


themoah:
Whenever there is a body in request env variable shows `Http_Content_Length=-1` (if empty shows 0).

## My actions before raising this issue
- [ ] Followed the [troubleshooting guide](https://docs.openfaas.com/deployment/troubleshooting/)
- [x] Read/searched [the docs](https://docs.openfaas.com/)
- [x] Searched [past issues](/issues)

## Expected Behaviour
I was expecting to see the length of body, especially for text (not binary)

## Current Behaviour
Always shows -1

## Possible Solution

Probably change https://github.com/openfaas/faas/blob/master/watchdog/handler.go#L238
to `%s` (though it's int64 type). I guess, I can submit bug fix bit later.

## Steps to Reproduce (for bugs)
1. `faas-cli deploy --name env --fprocess="env" --image="functions/alpine:latest"`
2. (if needed) port-forward your cli
3. `curl 'http://127.0.0.1:31112/function/env/' -X POST -d "" | grep -I length`  And see 0 (or `echo "" | faas-cli invoke env` 
4.  `curl 'http://127.0.0.1:31112/function/env/' -X POST -d "Alice " | grep -I length ` And see -1

## Your Environment

* FaaS-CLI version ( Full output from: `faas-cli version` ):
```
CLI:
 commit:  0ab4012784510381e2b90c031ede3d77197b10b1
 version: 0.11.0

Gateway
 uri:     http://127.0.0.1:31112
 version: 0.18.7
 sha:     59b7839236098820e73ed25301258b722c3d33e4
 commit:  Change how and when we fetch and parse namespace info


Provider
 name:          faas-netes
 orchestration: kubernetes
 version:       0.9.15
 sha:           41c33f9f7c29e8276bd01387f78d6f0cff847890
Your faas-cli version (0.11.0) may be out of date. Version: 0.11.3 is now available on GitHub.
```
* Docker version `docker version` (e.g. Docker 17.0.05 ): 19.03.5

* Are you using Docker Swarm or Kubernetes (FaaS-netes)? k8s on DO.

* Operating System and version (e.g. Linux, Windows, MacOS): Mac

* Code example or link to GitHub repo or gist to reproduce problem:

* Other diagnostic information / logs from [troubleshooting guide]

it's in classic-watchdog. Afaik everything goes toward of-watchdog and thus issue should be fixed there. 

## Next steps

[...]

PR #1423, opened 2020-01-05
Add transfer-encoding to watchdog env-vars if present

alexellis:

## Description

Add transfer-encoding to watchdog env-vars if present

## Motivation and Context
The Content-Type header must be ignored when the Transfer
Encoding is set to "chunked" because the length is unknown

Go sets this to -1 and we pass that onto the user:
https://golang.org/src/net/http/transfer.go

The value of Content_Length is currently set to -1 in this
scenario, however it caused some confusion for at least one
user in issue: #1422.

## How Has This Been Tested?

The Http_Transfer_Encoding value was tested by running the
watchdog on Linux with "env" as the fprocess and an extra
header to "curl"

## Types of changes

## Checklist:

### DIFF
--- a/watchdog/README.md
+++ b/watchdog/README.md
@@ -38,9 +38,9 @@ Here's how to package your function if you don't want to use the CLI or have exi
 Example Dockerfile for an `echo` function:
 
 ```
-FROM alpine:3.8
+FROM alpine:3.11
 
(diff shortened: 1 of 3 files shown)


codegold79:
I'll take a look at this and see what I can do.

PR #1430, opened 2020-01-14
Check err from NewRequest

yihuaf:
## Description

Error was ignored when creating a new request. When the endpoint url is
not welformed, ending without `/` for example, the `get` request will be
nil and client.Do will segfault.
## Motivation and Context

error out instead of segfault

## Types of changes

## Checklist:

### DIFF
--- a/gateway/metrics/exporter.go
+++ b/gateway/metrics/exporter.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"path"
 	"time"
 
 	"log"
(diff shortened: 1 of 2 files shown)


yihuaf:
Caching the request outside the loop is fine since the body is nil. I'd prefer to return an error up to the caller rather than log.Fatal on the spot.

As to forming the url, we can update the path using path.Join instead of string concat. We would need to check for error regardless though.

I will follow up with a new commit.

alexellis:
Taken a look at the error, it looks to be from the certifier -

alexellis:
I've got CI passing now, please look at my comment @yihuaf and thanks for your PR

yihuaf:
Updated based on comments.

yihuaf:
@alexellis PTAL with the new changes

yihuaf:
@LucasRoesler and @alexellis Please take a look at the updated PR based on previous feedback. Thanks.

gateway/server.go
```diff
@@ -62,7 +62,11 @@ func main() {
 
 	metricsOptions := metrics.BuildMetricsOptions()
 	exporter := metrics.NewExporter(metricsOptions, credentials)
-	exporter.StartServiceWatcher(*config.FunctionsProviderURL, metricsOptions, "func", servicePollInterval)
+	if err := exporter.StartServiceWatcher(*config.FunctionsProviderURL,
```
LucasRoesler: This line break is a little odd, it will read better on one line _or_ with each arg on a new line. In the second case the if block should follow the method call.

gateway/metrics/exporter.go
```diff
@@ -85,18 +86,24 @@ func (e *Exporter) StartServiceWatcher(endpointURL url.URL, metricsOptions Metri
...
 			case <-ticker.C:
-
-				get, _ := http.NewRequest(http.MethodGet, endpointURL.String()+"system/functions", nil)
-				if e.credentials != nil {
-					get.SetBasicAuth(e.credentials.User, e.credentials.Password)
-				}
-
 				services := []types.FunctionStatus{}
-				res, err := proxyClient.Do(get)
+				res, err := proxyClient.Do(serviceReq)
```
alexellis: I don't think we should be re-using the same HTTP request object, that seems wrong.

Why was that change made?

yihuaf: You mentioned in one of the previous comments that the request can be cached. From what I understand, a http request can be cached if the body is nil, which is the case here. I will make the change back to original if you think it's safer that way.

gateway/server.go
yihuaf: Done

yihuaf:
> LGTM, but shouldn't we also form a better URL and cache it outside of the loop/ticker?

> I don't think we should be re-using the same HTTP request object, that seems wrong.
> 
> Why was that change made?

> Not sure the change is correct.

You mentioned in one of the previous comments that the request should be cached. From what I understand, a http request can be cached if the body is nil, which is the case here. I can make the change back to original if you think it's safer that way. Let me know how you want me to proceed. @alexellis

PR #1431, opened 2020-01-16
Update ADOPTERS.md

matevzmihalic:
Signed-off-by: Matevz Mihalic <matevz.mihalic@gmail.com>

Include 3fs in adopters list.


alexellis:
It would be great if you could add the use-case like you told me on Slack, this is good, but too vague and the ADOPTERS file is trying to help new folks see how they can have success.

matevzmihalic:
I've added some use cases.

PR #1443, opened 2020-02-04
gateway/handlers: fix use of testing.T inside goroutine

alrs:
Here is a fix for `Test_logsProxyDoesNotLeakGoroutinesWhenClientClosesConnection()` in gateway/handlers`.

`testing.T.FatalF()` is a convenience function that calls `testing.T.Log()` and `testing.T.FailNow()`. From the `godoc` for `testing.T.FailNow()`:

> FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines.

This changes the test by creating an error channel to get the error variable outside of the goroutine, where it is checked after `cancel()`.

### DIFF
--- a/gateway/handlers/logs_test.go
+++ b/gateway/handlers/logs_test.go
@@ -122,13 +122,17 @@ func Test_logsProxyDoesNotLeakGoroutinesWhenClientClosesConnection(t *testing.T)
 		t.Fatalf("unexpected error sending log request: %s", err)
 	}
 
+	errCh := make(chan error, 1)
 	go func() {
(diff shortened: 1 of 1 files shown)


alexellis:
Hi, thanks for the PR. Paging @LucasRoesler PTAL

alexellis:
Thank you for the PR @alrs

alexellis:
You're now an OpenFaaS contributor and will appear on openfaas.com and at https://kenfdev.o6s.io/github-stats-page#/

PR #1474, opened 2020-03-11
Require PowerShell Desktop Edition for deploy_stack.ps1

arnydo:
## Description
Currently, .NET Core does not support System.Web.dll which causes `[System.Web.Security.Membership]::GeneratePassword(14, 5)` to fail.

When running the script with a `Core` edition of PowerShell such as `6.0.0` or `7.0.0` the script runs but fails to generate a password. The script continues to completion with no way to access the UI or CLI since the password is unknown/no-set.

Here is the output of the script running in `7.0.0`.

```powershell
PS>.\deploy_stack.ps1
InvalidOperation: C:\Working\faas\deploy_stack.ps1:30
Line |
  30 |      $password = [System.Web.Security.Membership]::GeneratePassword(24 …
     |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Unable to find type [System.Web.Security.Membership].

MethodInvocationException: C:\Working\faas\deploy_stack.ps1:36
Line |
  36 |          $hash = $sha256.ComputeHash([System.Text.Encoding]::UTF8.GetB …
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception calling "GetBytes" with "1" argument(s): "Array cannot be null. (Parameter 'chars')"

MethodInvocationException: C:\Working\faas\deploy_stack.ps1:38
Line |
  38 |          $secret = [System.BitConverter]::ToString($hash).Replace('-', …
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Exception calling "ToString" with "1" argument(s): "Value cannot be null. (Parameter 'value')"

Attempting to create credentials for gateway..
```

## Motivation and Context
The reason for the change is to provide a way to notify the end user that PowerShell `Core` is not (currently) supported and must be run with a `Desktop` edition (such as `5.1`).

Here is a reference to the issue regarding .Net Core not supporting `system.web.dll`: https://github.com/PowerShell/PowerShell/issues/5352

## How Has This Been Tested?
Tested the `build_stack.ps1` script in both `Core` and `Desktop` editions of PowerShell with expected results.

## Types of changes

## Checklist:

### DIFF
--- a/deploy_stack.ps1
+++ b/deploy_stack.ps1
@@ -1,4 +1,5 @@
 #!ps1
+#Requires -PSEdition Desktop
 
 param (
     [switch] $noAuth,


burtonr:
Hello @arnydo thank you for your interest! 

Please be sure to read the [Contribution Guide](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#i-have-a-great-idea) specifically at the section linked.

We prefer that you raise an issue first, so that we can agree on a plan before spending the time to make the change.

In this case, I think it would be a much better experience to update the PowerShell script to work across all platforms, rather than simply forcing the user to use a specific version.

Please raise an issue for this so we may discuss it further and come up with a proper plan

arnydo:
My apologies. I will create the issue.

PR #1488, opened 2020-03-22
Add namespace in function name for metrics

viveksyngh:
This commit adds namespace in function names while logging metrics to
prometheus, irrespective of the function is invoked with namespace suffix
or not.

This is also required to add multiple namespace support to faas-idler

https://github.com/openfaas-incubator/faas-idler/issues/37 which is part
of https://github.com/openfaas/faas-netes/issues/511

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/gateway/metrics/exporter.go
+++ b/gateway/metrics/exporter.go
@@ -23,17 +23,19 @@ import (
 
 // Exporter is a prometheus exporter
 type Exporter struct {
-	metricOptions MetricOptions
-	services      []types.FunctionStatus
-	credentials   *auth.BasicAuthCredentials
+	metricOptions     MetricOptions
+	services          []types.FunctionStatus
+	credentials       *auth.BasicAuthCredentials
+	FunctionNamespace string
 }
 
 // NewExporter creates a new exporter for the OpenFaaS gateway metrics
-func NewExporter(options MetricOptions, credentials *auth.BasicAuthCredentials) *Exporter {
+func NewExporter(options MetricOptions, credentials *auth.BasicAuthCredentials, namespace string) *Exporter {
 	return &Exporter{
-		metricOptions: options,
-		services:      []types.FunctionStatus{},
-		credentials:   credentials,
+		metricOptions:     options,
+		services:          []types.FunctionStatus{},
+		credentials:       credentials,
+		FunctionNamespace: namespace,
 	}
 }
 
@@ -91,8 +93,7 @@ func (e *Exporter) StartServiceWatcher(endpointURL url.URL, metricsOptions Metri
 				}
 
 				if len(namespaces) == 0 {
-					emptyNamespace := ""
-					services, err := e.getFunctions(endpointURL, emptyNamespace)
+					services, err := e.getFunctions(endpointURL, e.FunctionNamespace)
 					if err != nil {
 						log.Println(err)
 						continue
@@ -170,8 +171,9 @@ func (e *Exporter) getFunctions(endpointURL url.URL, namespace string) ([]types.
 
 func (e *Exporter) getNamespaces(endpointURL url.URL) ([]string, error) {
 	namespaces := []string{}
+	endpointURL.Path = path.Join(endpointURL.Path, "system/namespaces")
 
-	get, _ := http.NewRequest(http.MethodGet, endpointURL.String()+"system/namespaces", nil)
+	get, _ := http.NewRequest(http.MethodGet, endpointURL.String(), nil)
 	if e.credentials != nil {
 		get.SetBasicAuth(e.credentials.User, e.credentials.Password)
 	}
--- a/gateway/metrics/exporter.go
+++ b/gateway/metrics/exporter.go
@@ -6,10 +6,12 @@ package metrics
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
+	"path"
 	"time"
 
 	"log"
@@ -56,8 +58,14 @@ func (e *Exporter) Collect(ch chan<- prometheus.Metric) {
 
 	e.metricOptions.ServiceReplicasGauge.Reset()
 	for _, service := range e.services {
+		var serviceName string
+		if len(service.Namespace) > 0 {
+			serviceName = fmt.Sprintf("%s.%s", service.Name, service.Namespace)
+		} else {
+			serviceName = service.Name
+		}
 		e.metricOptions.ServiceReplicasGauge.
-			WithLabelValues(service.Name).
+			WithLabelValues(serviceName).
 			Set(float64(service.Replicas))
 	}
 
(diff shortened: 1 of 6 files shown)


gateway/metrics/exporter.go
```diff
@@ -72,9 +80,46 @@ func (e *Exporter) StartServiceWatcher(endpointURL url.URL, metricsOptions Metri
...
+					emptyNamespace := ""
+					services, err := e.getFunctions(endpointURL, emptyNamespace)
+					if err != nil {
+						log.Println(err)
+						continue
+					}
+					e.services = services
+				} else {
+					for _, namespace := range namespaces {
+						services, err := e.getFunctions(endpointURL, namespace)
```
LucasRoesler: I think this works well for now, but I also suspect a larger discussion is needed about how/if we move this into the provider sending a request per namespace every second seems like something the provider should encapsulate so that it can optimize that behavior as it best makes sense for itself.  

This current implementation would create a ["thundering herd"](https://en.wikipedia.org/wiki/Thundering_herd_problem)-like problem with no way to really streamline or optimize it.   For example, the k8s provider should be able to request the existing functions from several/all namespaces in a single request.  

Perhaps we need a new endpoint in the provider that allows the gateway to send 0, 1, or more namespaces as part of the request. When it is 0, it would request for all namespaces, otherwise the response would send back a list of functions for the provided namespaces. Alternatively, we could allow a special `_all` value. Then 0 namespaces == default, 1 or more applied as obvious, `_all` return functions from all possible namespaces.

gateway/metrics/exporter.go
```diff
@@ -87,40 +132,70 @@ func (e *Exporter) StartServiceWatcher(endpointURL url.URL, metricsOptions Metri
...
+	if unmarshalErr != nil {
+		return services, unmarshalErr
+	}
+	return services, nil
+}
+
+func (e *Exporter) getNamespaces(endpointURL url.URL) ([]string, error) {
+	namespaces := []string{}
+
+	get, _ := http.NewRequest(http.MethodGet, endpointURL.String()+"system/namespaces", nil)
```
LucasRoesler: Don't use string concatenation for path joining, it can be error prone in a way that you will only see at runtime.


This will build the same string safely everytime:
```suggestion
	endpointURL.Path = path.Join(endpointURL.Path, "system", "namespaces")
	get, _ := http.NewRequest(http.MethodGet, endpointURL.String(), nil)
```

gateway/metrics/exporter.go
```diff
@@ -72,9 +80,46 @@ func (e *Exporter) StartServiceWatcher(endpointURL url.URL, metricsOptions Metri
...
+			select {
+			case <-ticker.C:
+
+				namespaces, err := e.getNamespaces(endpointURL)
+				if err != nil {
+					log.Println(err)
+				}
+
+				if len(namespaces) == 0 {
+					emptyNamespace := ""
```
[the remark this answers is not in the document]
viveksyngh: Yes, I will add that

gateway/metrics/exporter.go
viveksyngh: I completely agree and we should move function aggregation logic to the provider. 

I think we can modify the exiting function list endpoint using some special query parameters  like `_all`, for which will combine functions from all namespaces and return it. Otherwise it works as it is working now. I think it will be a very minimal change compared to having a new endpoint. For new endpoint we will have to write different logic depending on the provider at several places.

gateway/metrics/exporter.go
viveksyngh: I will update this as well.

gateway/metrics/exporter.go
alexellis: Introducing a thundering herd issue concerns me. What are the alternatives to doing that?

gateway/metrics/exporter.go
```diff
@@ -72,9 +82,45 @@ func (e *Exporter) StartServiceWatcher(endpointURL url.URL, metricsOptions Metri
 	ticker := time.NewTicker(interval)
 	quit := make(chan struct{})
 
-	timeout := 3 * time.Second
+	go func() {
+		for {
```
alexellis, line 86: What is this new segment of code for? I was mainly expecting this PR to set a default suffix in the invocation URL when there was not one given.

viveksyngh: New segment of code is for getting replica count of the services, which was not having support for multiple namespace. It was only getting replica count for the functions present in default namespace.

gateway/metrics/exporter.go
viveksyngh: Thundering herd issue is because every 5 second (configurable) it tries to get all namespaces and then for each namespace it lists its functions to count service/replica count of functions. 

Solution is to use a new endpoint or existing which list all functions across all namespaces so that we don't have to make these multiple calls. This implementation can be done in faas-netes which will be efficient.

alexellis:
This was reviewed several weeks ago in a members call with @viveksyngh and @LucasRoesler - we had a concern about performance and paused the patch. We need to review the solution and see if there are alternatives with other trade-offs (this still may be the best option)

viveksyngh:
Testing :

```
k3d create 

export KUBECONFIG="$(k3d get-kubeconfig --name='k3s-default')"

arkade install openfaas --set clusterRole=true



kubectl port-forward -n openfaas svc/gateway 8080:8080 &

PASSWORD=$(kubectl get secret -n openfaas basic-auth -o jsonpath="{.data.basic-auth-password}" | base64 --decode; echo)

echo -n $PASSWORD | faas-cli login --username admin --password-stdin

kubectl -n openfaas run \
--image=stefanprodan/faas-grafana:4.6.3 \
--port=3000 \
grafana

kubectl -n openfaas expose deployment grafana \
--type=NodePort \
--name=grafana

kubectl port-forward -n openfaas svc/grafana 3000:3000 &

kubectl port-forward -n openfaas svc/prometheus 9090:9090 &

kubectl create namespace fn

kubectl annotate ns/fn openfaas=1

faas-cli namespaces
Namespaces:
 - fn
 - openfaas-fn



faas-cli store deploy figlet 

faas-cli store deploy figlet -n fn

faas-cli store deploy nodeinfo -n fn

faas-cli store deploy leftpad

echo "Hello" | faas-cli invoke figlet -n fn

echo "Hello" | faas-cli invoke figlet

echo "Hello" | faas-cli invoke leftpad

echo "Hello" | faas-cli invoke nodeinfo -n fn


➜  ~ faas-cli list -n fn
Function                      	Invocations    	Replicas
figlet                        	1              	1
nodeinfo                      	0              	1
➜  ~ faas-cli list
Function                      	Invocations    	Replicas
figlet                        	1              	1
leftpad                       	1              	1
```

* Before Changes 

Screen Shot 2020-07-29 at 3 31 13 PM
Screen Shot 2020-07-29 at 3 31 04 PM
Screen Shot 2020-07-29 at 3 30 44 PM

* After Changes

Edit the Gateway deployment to use docker image `viveksyngh/gateway:metrics-with-namespace-1`

```
faas-cli list -n openfaas-fn
Function                      	Invocations    	Replicas
figlet                        	2              	1
leftpad                       	1              	1
➜  ~ faas-cli list -n fn
Function                      	Invocations    	Replicas
nodeinfo                      	5              	1
figlet                        	3              	1
```

Screen Shot 2020-07-29 at 4 17 58 PM
Screen Shot 2020-07-29 at 4 17 39 PM
Screen Shot 2020-07-29 at 4 17 27 PM

gateway/metrics/exporter.go
alexellis: @viveksyngh can you write a bash loop to deploy 100-200 functions and then measure the effect of the patch?

alexellis:
```bash
for i in {1..100} do ;
   faas-cli deploy --image figlet --name fn$i
done
```

gateway/metrics/exporter.go
```diff
@@ -87,40 +133,71 @@ func (e *Exporter) StartServiceWatcher(endpointURL url.URL, metricsOptions Metri
...
+	}
+
+	unmarshalErr := json.Unmarshal(bytesOut, &services)
+	if unmarshalErr != nil {
+		return services, unmarshalErr
+	}
+	return services, nil
+}
+
+func (e *Exporter) getNamespaces(endpointURL url.URL) ([]string, error) {
```
alexellis, line 172: Can we use the CLI's code for this?

gateway/metrics/exporter.go
```diff
@@ -87,40 +133,71 @@ func (e *Exporter) StartServiceWatcher(endpointURL url.URL, metricsOptions Metri
 			ExpectContinueTimeout: 1500 * time.Millisecond,
 		},
 	}
+}
 
-	go func() {
-		for {
-			select {
-			case <-ticker.C:
+func (e *Exporter) getFunctions(endpointURL url.URL, namespace string) ([]types.FunctionStatus, error) {
```
alexellis, line 138: Can you use the CLI here instead?

alexellis:
@viveksyngh we will need a rebase on this.

After some discussion with @LucasRoesler, we thought this PR should be merged and then load-tested later.

alexellis:
@LucasRoesler could you give this a final pass?

LucasRoesler (approved):
@alexellis I think we are ready. Let's follow up in a future PR to consolidate those places we can/should use the cli

alexellis (approved):
Approved

PR #1551, opened 2020-07-06
Update ADOPTERS.md

mjallday:
Added VGS adoption statement. Related to #776 

If a logo is required there are such available at https://www.verygoodsecurity.com/press-and-assets

Happy to take feedback on the usage statement and adjust as we publish more details in the future about how we're integrating.

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -80,6 +80,8 @@ This is a list of adopters of OpenFaaS
 
 * [UStore](http://ustore.com.br/) - "We're using OpenFaaS in production"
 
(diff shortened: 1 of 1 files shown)


alexellis:
> Happy to take feedback on the usage statement and adjust as we publish more details in the future about how we're integrating.

Thank you for the PR and I think this is consistent with what other companies have put forward. What I'd be interested in doing is perhaps a segment for use in a blog post or conference talk later on - we could do that over email perhaps?

Thanks again for getting this done.

Alex

PR #1568, opened 2020-09-11
Fixes a type in `gateway/README.md`

alexgorbatchev:
Fixes a typo. Took me a while to figure out why port config wasn't working because I copied and pasted the value :(


alexellis:
Hi @alexgorbatchev can you say who are you working for? Our contribution guidelines state that users should log issues for typos, not raise PRs. PRs need to be committed with `git commit -s` as is documented and mentioned in the PR description guide above.

I would caution you on using Docker Swarm, and advise that you pick up faas-netes for Kubernetes instead which is documented along with its helm chart here -> https://github.com/openfaas/faas-netes/tree/master/chart/openfaas

PR #1655, opened 2021-06-24
adopters smashHit.eu added

tekrajchhetri:
We are using OpenFaaS in our project and updated adopters list


alexellis:
Thank you for the addition.

Would you like to fix up the PR using the command you were given by Derek, or for us to update the file on your behalf?

alexellis:
Thank you :+1:

PR #1685, opened 2021-12-20
feat: add OTEL tracing span middleware during function proxy

### LINKED ISSUE #1684: Add OpenTelemetry support during function proxy
## My actions before raising this issue
- [ ] Followed the [troubleshooting guide](https://docs.openfaas.com/deployment/troubleshooting/)
- [ ] Read/searched [the docs](https://docs.openfaas.com/)
- [x] Searched [past issues](/issues)

## Expected Behaviour

During function proxy, the Gateway should be able to produce open telemetry spans.

## Current Behaviour

There are no tracing spans

## Which Solution Do You Recommend?

I recently did a walk-through for integrating OpenTelemetry with OpenFaaS functions and think it would be nice if the Gateway could produce an OpenTelemetry spans during function invocation.  Adding tracing during the function proxy would provide a more accurate picture of the networking in the cluster and enable accurate assessments of the overhead (or lack thereof) from the Gateway.

We previously discussed this in general in https://github.com/openfaas/faas/issues/1354 but OpenTelemetry was not a active project at the time, only OpenTracing. OpenTelemetry. OpenTelemetry makes this integration much more feasbile now because we can more easily provide support for multiple exporters. Additionally, the OpenTelemetry providers generally allow all of the required configuration via env variables, which means the integration should require only minimal changes to the Gateway.

During the Gateway startup we would initialize and set the global tracing provider using something like this 

```go
shutdownTracing, err := tracing.Provider(config.Version, config.Commit)
if err != nil {
	log.Fatal(err)
}
// Cleanly shutdown and flush telemetry when the application exits.
defer shutdownTracing(ctx)
```

We can then encapsulate all of the tracing specific code in the `Provider` implemenation

```go
func Provider(version, commit string) (shutdown Shutdown, err error) {
	exporter := Exporter(os.Getenv("OTEL_EXPORTER"))

	var exp tracesdk.TracerProviderOption
	switch exporter {
	case JaegerExporter:
		// configure the collector from the env variables,
		// OTEL_EXPORTER_JAEGER_ENDPOINT/USER/PASSWORD
		j, e := jaeger.New(jaeger.WithCollectorEndpoint())
		exp, err = tracesdk.WithBatcher(j), e
	case LogExporter:
		w := os.Stdout
		opts := []stdouttrace.Option{stdouttrace.WithWriter(w)}
		if truthyEnv("OTEL_EXPORTER_LOG_PRETTY_PRINT") {
			opts = append(opts, stdouttrace.WithPrettyPrint())
		}
		if !truthyEnv("OTEL_EXPORTER_LOG_TIMESTAMPS") {
			opts = append(opts, stdouttrace.WithoutTimestamps())
		}

		s, e := stdouttrace.New(opts...)
		exp, err = tracesdk.WithSyncer(s), e
	// additional exporters
	default:
		logrus.Warn("tracing disabled")
		// We explicitly DO NOT set the global TracerProvider using otel.SetTracerProvider().
		// The unset TracerProvider returns a no-op "non-recording" span, but still passes through context.
		otel.SetTextMapPropagator(
			propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}),
		)
		// return no-op shutdown function
		return func(_ context.Context) {}, nil
	}
	if err != nil {
		return nil, err
	}
	
	// some additional work to
	// finish initializing the provider 
	
	otel.SetTracerProvider(provider)

	shutdown = func(ctx context.Context) {
		// Do not let the application hang forever when it is shutdown.
		ctx, cancel := context.WithTimeout(ctx, time.Second*5)
		defer cancel()

		err := provider.Shutdown(ctx)
		if err != nil {
			logrus.WithError(err).Error("tracing provider did not gracefully shutdown")
		}
	}
	return shutdown, nil
}
```

Inside the function invocation hanlder here https://github.com/openfaas/faas/blob/8a87b57ce124d909fa8220259bdaf13eb6a152a4/gateway/handlers/forwarding_proxy.go#L55 we would add

```go
	var err error
	_, span := otel.Tracer("Gateway").Start(r.Context(), "Proxy")
	defer func() {
		if err != nil {
			span.SetStatus(codes.Error, err.Error())
			span.RecordError(err)
		}
		span.End()
	}()
```

This would then show as a new span named "Proxy" between the ingress and the function (if they have tracing enabled).  There are a few other things we could do, e.g. adding the status code, original url, and request url as metadata to the span, but this is optional for a minimal implementation. 

1. Follow this walkthrough https://github.com/LucasRoesler/openfaas-tracing-walkthrough

## Context

https://github.com/LucasRoesler/openfaas-tracing-walkthrough

LucasRoesler:
## Description

Add tracing package the standardizes an OpenTelemetry tracing initialization and Middleware that can be applied to the Gateway handlers. 

Add the tracing middleware to the function proxy handler only. 

Requires new env variables to be enabled, it is otherwise a no-op.

## Motivation and Context

Resolves #1684 

## How Has This Been Tested?

Manually tested by using the tracing walkthrough https://github.com/LucasRoesler/openfaas-tracing-walkthrough but modifying the Gateway image to a locally built image from this branch

## Types of changes

## Checklist:


gateway/main.go
```diff
@@ -154,6 +162,7 @@ func main() {
 		scaler := scaling.NewFunctionScaler(scalingConfig, scalingFunctionCache)
 		functionProxy = handlers.MakeScalingHandler(faasHandlers.Proxy, scaler, scalingConfig, config.Namespace)
 	}
+	functionProxy = tracing.Middleware(tracing.ConstantName("FunctionProxy"), functionProxy)
```
LucasRoesler, line 165: this applies the tracing middleware to only the function proxy, we chould also choose to add it to other endpoints though

gateway/pkg/tracing/middleware.go
```diff
@@ -0,0 +1,62 @@
...
+
+// ConstantName geneates the given name for the span based on the request.
+func ConstantName(value string) func(*http.Request) string {
+	return func(r *http.Request) string {
+		return value
+	}
+}
+
+func debug(span trace.Span, format string, args ...interface{}) {
+	value := os.Getenv("OTEL_LOG_LEVEL")
```
LucasRoesler, line 56: this is the standard env variable for the otel tooling log level, but it isn't fully respected by the go implementation yet. This was helpful for debugging some early issues with the middleware

gateway/pkg/tracing/otel.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+type Shutdown func(context.Context)
+
+// Provider returns an OpenTelemetry TracerProvider configured to use
+// the Jaeger exporter that will send spans to the provided url. The returned
+// TracerProvider will also use a Resource configured with all the information
+// about the application.
+func Provider(ctx context.Context, name, version, commit string) (shutdown Shutdown, err error) {
+	exporter := Exporter(os.Getenv("OTEL_EXPORTER"))
```
LucasRoesler, line 39: This is not a standard env variable, we could choose a new name for it if you want.

gateway/pkg/tracing/otel.go
```diff
@@ -0,0 +1,155 @@
...
+		if !truthyEnv("OTEL_EXPORTER_LOG_TIMESTAMPS") {
+			opts = append(opts, stdouttrace.WithoutTimestamps())
+		}
+
+		s, e := stdouttrace.New(opts...)
+		exp, err = tracesdk.WithSyncer(s), e
+	case OTELExporter:
+		// find available env variables for configuration
+		// see: https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/otlp/otlptrace#environment-variables
+		kind := get("OTEL_EXPORTER_OTLP_PROTOCOL", "grpc")
```
LucasRoesler, line 64: this is also not standard, but  i wanted a way to control which client the exporter used

berylshow:
Is there any progress in this task? thank you

LucasRoesler:
@berylshow it is currently stuck, @alexellis has asked the community for some input on use cases to gauge interest in the feature before we merge it.

